Issue #125, opened 2026-08-07
Type-check tests/ with mypy (annotations are currently unverified)


MehdiZare:
## Problem

`pyproject.toml` `[tool.mypy]` excludes `tests/`:

```toml
exclude = [
    "tests/",
    "examples/",
    "scripts/",
    "build/",
    "dist/",
]
```

The project's stated convention is "Type hints are required" (CLAUDE.md), and CodeRabbit enforces it in review — see the outside-diff finding on #113 that produced #124. But because `tests/` is excluded, **every annotation we add to a test file is documentation-only and machine-unverified**. Nothing catches an annotation that is wrong, drifts, or contradicts the fixture it describes.

That is the worst of both worlds: we pay the review cost of demanding annotations without getting the guarantee they exist to provide.

## Evidence of scale

Running mypy directly against just the three files touched by #124, with the `tests/` exclude bypassed:

```bash
uv run mypy --disallow-untyped-defs --disallow-incomplete-defs --check-untyped-defs \
  tests/unit/test_intelligence.py \
  tests/unit/test_async_clients.py \
  tests/integration/test_intelligence.py
```

→ **266 errors**, all pre-existing and all on lines outside the ranges #124 touched. So flipping the exclude on is not a one-line change; three files alone carry 266 latent errors.

## Proposed approach

Incremental, so it never blocks a feature PR:

1. Remove `"tests/"` from `[tool.mypy].exclude`.
2. Add a per-module override that relaxes the strict flags for `tests.*` initially, so CI stays green:
   ```toml
   [[tool.mypy.overrides]]
   module = ["tests.*"]
   disallow_untyped_defs = false
   disallow_incomplete_defs = false
   ```
   This immediately catches *wrong* annotations (the actual risk) without demanding *complete* ones.
3. Burn down file-by-file, tightening the override scope as directories are cleared. Suggested order by churn: `tests/unit/`, then `tests/integration/`, then the rest.
4. Once clear, drop the override entirely so `tests/` inherits the project's strict settings.

Step 2 alone captures most of the value and should be small.

## Non-goals

- Not a full test-suite typing sweep in one PR.
- Not changing any test behavior or assertions.

## Context

- Surfaced during code review of #124 (test annotation polish for release #113).
- #124 deliberately scoped itself to the exact ranges CodeRabbit flagged and did **not** expand into a full-file sweep; this issue tracks the broader fix.

Refs: #113, #124

MehdiZare:
[...]

Issue #123, opened 2026-08-06
LC/MCP: remaining semantics parameter-hint and catalog-count drift


MehdiZare:
## Context

Broader cleanup noted while landing #120.

### Parameter hint drift

Several endpoints fail LC `_validate_parameters` because optional/mandatory params lack matching `parameter_hints` (examples: `get_equity_offering_rss` missing `cik`, `get_market_hours` missing `exchange`, `get_form_13f` missing year/quarter, `get_analyst_estimates` missing page/limit/period).

### Docs count conventions

`docs/mcp/tools.md` section headers often list full semantics/catalog counts (e.g. Intelligence 45) while `DEFAULT_TOOLS` is smaller (39). Pre-existing and intentional-ish, but easy to misread. Consider documenting the convention once, or listing both numbers.

### Undocumented MCP sections

PR #120 notes pre-existing omission of `batch`, `index`, `sec`, `transcripts`, and `company.profile_cik` (~52 tools) from `docs/mcp/tools.md`.

### Inconsistent institutional naming

- Endpoint `name`: `cik_mapper` / `cik_mapper_by_name`
- Client methods: `get_cik_mappings` / `search_cik_by_name`
- MCP semantics keys: `cik_mappings` / `cik_mapper_by_name`

Not wrong for MCP after #120, but worth a naming cleanup pass.

Related: #114, #115, #116, PR #120.

MehdiZare:
Concrete scope from the PR #120 review pass, so this issue can be worked without re-deriving it.

**Finding:** `EndpointSemantics.response_hints` keys are never validated against the response model's fields. They only feed the LangChain vector-store embedding string (`f"response {field_name}:"` in `fmp_data/lc/registry.py`), so a key naming no real field silently degrades semantic search rather than failing.

**Current drift in `fmp_data/intelligence/mapping.py`** — 10 entries with hint keys that match no field on their endpoint's `response_model` (measured on `dev` + #120, after #120 fixed the one entry it introduced):

| Semantics key | Unknown hint key(s) |
|---|---|
| `earnings_calendar` | `date` |
| `historical_earnings` | `date` |
| `dividends_calendar` | `date` |
| `stock_splits_calendar` | `date` |
| `equity_offering_rss` | `offering_type`, `amount` |
| `equity_offering_search` | `offering_type`, `amount` |
| `equity_offering_by_cik` | `offering_amount` |
| `crowdfunding_by_cik` | `offering_details` |
[...]

PR #120, opened 2026-08-06
fix(mcp): repair intelligence mapping drift and wire grades/ratings tools (#114, #115, #116)

MehdiZare:
Closes #114, closes #115, closes #116, closes #122.

## #114 — mapping drift between semantics and client methods

`register_from_manifest` resolves `client.<method_name>`, so a semantics entry whose `method_name` no client implements fails at registration even though discovery advertises it.

- `intelligence`: `get_crowdfunding_search` → `search_crowdfunding`, `get_equity_offering_search` → `search_equity_offering`. `INTELLIGENCE_ENDPOINT_MAP` keys renamed to match, since those keys *are* method names (`fmp_data/lc/registry.py` and the vector store look endpoints up by `semantics.method_name`). Tool keys (`crowdfunding_search`, `equity_offering_search`) are unchanged.
- The new guard test surfaced the same defect class in two neighbouring clients, fixed here so the invariant can be enforced repo-wide rather than intelligence-only:
  - `market.search` → `search_company`
  - `institutional`: `get_cik_mapper` → `get_cik_mappings` (semantics key `cik_mapper` → `cik_mappings`), `get_cik_mapper_by_name` → `search_cik_by_name`

Broken tools before: 7. After: 0 (of 224 discovered).

## #115 — ghost intelligence semantics

Removed `intelligence.institutional_holders` and `intelligence.financial_reports_dates`, which advertised methods living on the institutional and fundamental clients. They could never register, and their key-only tool names collided with the real `institutional.institutional_holders` / `fundamental.financial_reports_dates` tools. A comment in `mapping.py` records why they must not come back.

## #116 — grades/ratings surface

Added `INTELLIGENCE_ENDPOINT_MAP` entries and `INTELLIGENCE_ENDPOINTS_SEMANTICS` for: `ratings_snapshot`, `ratings_historical`, `price_target_news`, `price_target_latest_news`, `grades`, `grades_historical`, `grades_consensus`, `grades_news`, `grades_latest_news`.

All nine added to `DEFAULT_TOOLS`, along with `intelligence.crowdfunding_search` and `intelligence.equity_offering_search`, which can now actually register. Intelligence defaults: 28 → 39.

All nine paths were re-verified against the live FMP API during review — every one returns HTTP 200 with a populated payload:

| path | status |
|---|---|
| `ratings-snapshot` | 200 |
| `ratings-historical` | 200 |
| `price-target-news` | 200 |
| `price-target-latest-news` | 200 |
| `grades` | 200 |
[...]

Issue #126, opened 2026-08-07
MCP: key-only tool specs 'crypto_quotes' and 'forex_quotes' are ambiguous across clients


MehdiZare:
Discovered while reviewing #120.

`register_from_manifest` accepts key-only tool specs (`"profile"` as well as `"company.profile"`) and resolves them through `_build_key_to_spec` (`fmp_data/mcp/tool_loader.py:50`), which indexes **all** discovered tools — not just `DEFAULT_TOOLS`.

Two keys are claimed by two clients each:

```
crypto_quotes -> ["alternative.crypto_quotes", "batch.crypto_quotes"]
forex_quotes  -> ["alternative.forex_quotes",  "batch.forex_quotes"]
```

## Repro

```python
from unittest.mock import Mock
from fmp_data.client import FMPDataClient
from fmp_data.mcp.tool_loader import register_from_manifest

register_from_manifest(Mock(), FMPDataClient(api_key="x"), ["crypto_quotes"])
# RuntimeError: Tool key 'crypto_quotes' is ambiguous
```

Anyone writing a key-only manifest for these two endpoints hits a hard failure with no way to disambiguate other than switching that entry to the full `<client>.<key>` form.

## Scope

Pre-existing; not introduced by #120. Neither key is in `DEFAULT_TOOLS`, so the default server is unaffected.

`tests/unit/test_mcp.py::TestSemanticsMethodResolution::test_discovered_tool_keys_are_unambiguous_globally` (added in #120) enforces that **no new** collisions appear, with these two on a shrinking allowlist. Closing this issue means emptying that allowlist.

## Options

1. Rename one side of each pair (e.g. `batch.crypto_quotes` -> `batch.crypto_quotes_bulk`) — changes a public MCP tool key.
2. Give key-only resolution a documented precedence order instead of raising.
3. Drop key-only resolution for ambiguous keys and require the full spec, with a clearer error naming both candidates.

MehdiZare:
Fixed by #147, merged to `dev` as `0d04519`.

Resolved via option 3: bare-key resolution is dropped for ambiguous keys and the error now names every candidate and the exact form to use. Neither tool is removed or renamed — what changed is the stated guarantee: **a bare tool key resolves only when exactly one client claims it.**

```
Tool key 'crypto_quotes' is claimed by 2 clients (alternative.crypto_quotes,
batch.crypto_quotes), so the bare key cannot be resolved. Bare tool keys work
[...]

PR #127, opened 2026-08-07
fix(lc): resilient endpoint registration, semantics drift, docs sync, and typed tests (#121, #122, #123, #125)

### LINKED ISSUE #121: LC registry: register_batch aborts whole client group on first validation error
## Context

Surfaced while reviewing #120 (MCP intelligence mapping drift / grades-ratings).

`setup_registry` builds a per-client batch and calls `EndpointRegistry.register_batch`. That method raises on the **first** invalid endpoint, and `setup_registry` then skips the **entire** client group.

### Observed examples (pre-existing)

- **intelligence** aborts on `get_equity_offering_rss` (endpoint has `cik` param; semantics missing `cik` hint) → none of the new grades/ratings endpoints land in the live registry path even when they validate alone.
- **company** aborts on `get_analyst_estimates` (missing page/limit/period hints).
- **market** aborts on `get_market_hours` (missing `exchange` hint).
- **institutional** aborts on `get_form_13f` (missing year/quarter hints).

### Why it matters

Even after #120 pairs endpoint-map keys correctly (including `search_*` via `method_name`), LangChain vector-store / registry consumers still miss large surfaces of the catalog.

### Proposed direction

1. Make registration resilient: register valid endpoints, collect/log failures, do not drop the whole group.
2. Optionally soft-validate parameter hints (warn vs hard fail) for optional params.
3. Add a CI guard that reports unpaired or validation-failed endpoints without requiring a full vector store.

Related: #114, #115, #116, PR #120.

### LINKED ISSUE #122: LC validation: price_target* prefix rules collide with intelligence price-target news tools
## Context

From PR #120 review (#116 grades/ratings surface).

New intelligence tools:

- `get_price_target_news`
- `get_price_target_latest_news`

can fail LangChain category validation because company-side rules claim the `price_target*` prefix for **Company Information**, while these endpoints are declared under **Intelligence**.

### Expected

Either:

- prefix rules become more specific (e.g. exact `get_price_target` / `get_price_target_summary` / `get_price_target_consensus` only), or
- intelligence price-target news tools are explicitly allowed under Intelligence.

### Out of scope for #120

MCP registration does not use the LC category validator; this only affects `setup_registry` / vector store paths.

Related: #116, PR #120.

MehdiZare:
Closes #121, #123, #125. Resolves #122 (see below).

## Why

`EndpointRegistry.register_batch()` raised on the first endpoint that failed validation, and `setup_registry()` caught that per group — so one drifted endpoint silently dropped the rest of its client group. Measured on `dev`: **75 of 168 endpoints** reached the live registry. Chasing that down surfaced the semantics drift causing the failures and a docs reference missing four whole client sections.

### #121 — registry no longer aborts a group
- `register_batch()` registers every valid endpoint, collects failures, and returns `{endpoint_name: error}` instead of raising
- `setup_registry()` logs a per-group summary of what it skipped
- `register()` no longer double-logs a validation failure with a traceback
- Non-validation errors now propagate out of `setup_registry()` instead of being swallowed — a group that cannot load at all should not resemble a group with nothing in it

### #123 — semantics drift (27 failures → 0)
- Added missing parameter hints: `analyst_estimates`, `intraday_prices`, `employee_count`, `geographic_revenue_segmentation`, `owner_earnings`, `market_hours`, `search_by_cik`, the sector/industry performance + PE endpoints, `form_13f`, `institutional_holders`, `institutional_holdings`, `insider_trades`, `cik_mapper*`, `equity_offering_rss`
- Removed hints for parameters the endpoints do not accept (`stock_news.tickers`, `stock_news_sentiments` date/limit, `institutional_holdings.includeCurrentQuarter`, `cik_mapper_by_name.name`) — hints are consumed by iterating endpoint params, so these were dead weight
- `aftermarket_trade`, `aftermarket_quote`, `stock_price_change` recategorized to Company Information, matching the client they live on
- Historical sector/industry endpoints now hint the real `from`/`to` query params instead of `start_date`/`end_date`
- New shared hints in `fmp_data.lc.hints`: `PAGE_HINT`, `SECTOR_HINT`, `INDUSTRY_HINT`, `YEAR_HINT`, `QUARTER_HINT`, `CIK_HINT`, `FROM_TO_DATE_HINTS`

### Found during review of this PR
- **`company.intraday_price` shipped with three undocumented parameters.** The alias shares `get_intraday_prices` with the `intraday_prices` entry but hinted only `symbol`/`interval`. It is in `DEFAULT_TOOLS`, and registry validation never saw it because the endpoint-map key pairs with `intraday_prices` first. Fixed, plus a semantics-first guard for alias entries no endpoint key selects.
- **Eight company semantics claimed `client_name="market"`** (`quote`, `simple_quote`, `intraday_price(s)`, `historical_price(s)`, `market_cap`, `historical_market_cap`) — those methods live on `CompanyClient`. Metadata-only (nothing dispatches on the field), now guarded.
- **The #121 regression guard was not running in the core CI session.** `tests/unit/lc/conftest.py` `importorskip`s langchain and faiss for the whole directory; neither guard file imports either. Moved to `tests/unit/`. Verified in a core-only venv: 7 tests that previously skipped now run.
- **The docs guard tolerated deleting a row.** It compared bare tool keys, and `crypto_quotes`/`forex_quotes` exist under two clients each, so 224 entries collapsed to 222 unique keys. Now compares fully-qualified `client.key`, checks both directions, and asserts all 13 per-section counts rather than only the two header totals.
- `InstitutionalOwnershipDates.cik` is now a declared optional field instead of relying on extra-field passthrough — the API returns it and a test asserted on it.

### #123 (docs) — complete + self-checking reference
- `docs/mcp/tools.md` gains the missing `Batch` (30), `Index` (6), `SEC` (12), `Transcripts` (4) sections and `company.profile_cik`
- Every section states both numbers — catalog tools vs `DEFAULT_TOOLS` — with the convention explained once in the header (224 catalog / 159 default), which is the confusion #123 flagged

### #125 — tests/ is type-checked
- Removed `tests/` from `[tool.mypy].exclude`; added a `tests.*` override relaxing `disallow_untyped_defs` / `disallow_incomplete_defs` so *wrong* annotations are caught without demanding *complete* ones
- Fixed the **76** errors this surfaced: narrowing asserts for Optional/union values, explicit annotations where a `Mock` left a TypeVar unbound, targeted `type: ignore` for deliberate invalid constructions and method patching, removal of stale ignores. **No test behavior or assertions changed.**
- `nox -s typecheck` now covers `fmp_data` and `tests`
- Remaining per the issue: tighten the override directory by directory — deliberately out of scope

### #122 — not reproducible, now guarded
The `price_target*` prefix collision does not occur: after #120, `get_price_target_news` and `get_price_target_latest_news` exist, resolve under Intelligence, and pass category validation. The new catalog guard test locks that in. Closing as resolved rather than fixing something that is not broken.

## Result

**168 of 168 endpoints** register cleanly (was 75).

## Corrections to the original description

Review caught three numbers that were wrong here and in the CHANGELOG, now fixed in both:

- **"73 of 159"** → the real base is **75 of 168**. `159` is `len(DEFAULT_TOOLS)`, an MCP number unrelated to the LangChain registry; the registry denominator is 168 and the MCP catalog is 224.
- **"29 failures"** → **27**. The 73/29 pair was measured before #120 added the `method_name` fallback to `resolve_semantics_for_endpoint`; exactly two endpoints pair only via that fallback, which accounts for both deltas.
- **"`market.search` was a broken MCP tool"** — it was not, and this PR does not touch it. On `dev`, `MARKET_ENDPOINTS_SEMANTICS["search"].method_name` is already `search_company`, `_resolve_attr` resolves it, and `market.search` is not in `DEFAULT_TOOLS`. That fix shipped in #120; the bullet double-counted it and has been removed.

Both base numbers were re-derived by replaying the old abort-on-first-failure path against an `origin/dev` tree and the new path against `HEAD`.

## Follow-ups filed rather than scope-creeping this PR

- #128 — LangChain tools mark every endpoint parameter as required, including optional ones. Pre-existing, but this PR takes the registry from 75 to 168 endpoints, so the blast radius more than doubled.
- #129 — `batch`, `sec` and `transcripts` semantics are unguarded and carry 33 missing hints. Fixing needs them added to `_get_endpoint_groups()` and the hints filled in together, or CI goes red in between.
- #130 — `institutional.cik_mapper_by_name` is a duplicate tool that cannot accept a name. Removing the `name` hint here was correct against the endpoint, but the entry needs a design decision.
- #131 — `InstitutionalOwnershipDates.cik` rejects an integer CIK in every validation mode. Needs a live-API probe to settle.

## Verification

- `1090 passed, 3 skipped` (21 new tests)
- `ruff check` + `ruff format --check` clean
- `mypy fmp_data tests` clean, 190 files
- Every new guard test verified non-vacuous by introducing the drift it is meant to catch and confirming the specific failure message

https://claude.ai/code/session_01KbwQTKm67bV4RHL1GNF8wN


MehdiZare:
@coderabbitai review

MehdiZare:
## Review round: findings and how each was handled

Two independent reviewers went over `origin/dev...HEAD` — one full-diff, one auditing the semantics/mapping claims by introspecting the live objects rather than reading the diff. Every claim below was re-verified before acting on it. Nine findings, seven fixed here as isolated commits, four filed as issues.

### Fixed in this PR

| # | Finding | Commit |
|---|---|---|
| 1 | **CHANGELOG counts wrong in both numbers.** "73 to the full 159" — `159` is `len(DEFAULT_TOOLS)`, an MCP number unrelated to the LC registry. Re-derived by replaying the old abort-on-first-failure path against an `origin/dev` tree and the new path against `HEAD`: **75 → 168**, and 27 invalid endpoints, not 29. | `4de8f2c` |
| 2 | **CHANGELOG double-counted #120.** The "broken `market.search` MCP tool" bullet describes work already on `dev` — `method_name` is already `search_company` there, `_resolve_attr` resolves it, and `market.search` is not in `DEFAULT_TOOLS`. Nothing in this diff touches it. Removed. | `4de8f2c` |
| 3 | **`register_batch()` changed its public contract under `### Fixed`.** `-> None` + raises became `-> dict[str, str]` + never raises. Moved to `### Changed` with the actual blast radius stated (`EndpointRegistry` is not in `fmp_data.lc.__all__`). | `4de8f2c` |
| 4 | **`company.intraday_price` shipped three undocumented parameters.** Real bug this PR would have missed. The alias shares `get_intraday_prices` with `intraday_prices` but hinted only `symbol`/`interval`; the endpoint also takes `start_date`, `end_date`, `nonadjusted`. It is in `DEFAULT_TOOLS`. Validation never saw it because the guard iterates endpoint-map-first and the endpoint key pairs with `intraday_prices`. Fixed + semantics-first guard. | `fc571cb` |
| 5 | **Eight company semantics declared `client_name="market"`** for methods on `CompanyClient`. Metadata-only — nothing dispatches on the field, which is why validation was blind to it. Fixed + guard. | `13692cb` |
| 6 | **The #121 regression guard was not running in the core CI session.** `tests/unit/lc/conftest.py` `importorskip`s langchain/faiss for the whole directory; neither guard file imports either. Moved to `tests/unit/`. Verified in a core-only venv: 7 tests that previously skipped now run. This is precisely the silent-disappearance failure mode the PR exists to prevent. | `f1707e1` |
| 7 | **The docs guard tolerated deleting a row.** It compared bare keys, and `crypto_quotes`/`forex_quotes` exist under two clients each, so 224 entries collapsed to 222 unique keys — deleting either row still passed. Now compares `client.key`, checks both directions, and asserts all 13 per-section counts instead of only the two header totals. | `d92ebe2` |
| 8 | Dead `conftest` mypy override (subsumed by `tests.*`); integration-test comment claiming an empty list "has been observed" when the adjacent `type: ignore[unreachable]` is mypy proving that branch impossible. | `c4b76c2` |
| 9 | `setup_registry()` losing its `except Exception` is a user-visible change on an exported function. Kept fail-loud — that is the right call — but written down. | `1d3abfe` |

Every new guard was verified non-vacuous by introducing the drift it targets and confirming the specific failure. E.g. reverting the `intraday_price` hunk fails with `missing=['end_date', 'nonadjusted', 'start_date']`; deleting the batch `crypto_quotes` row fails with `['batch.crypto_quotes']`.

### Filed as issues rather than widening this PR

- **#128** — LangChain tools mark every endpoint parameter as required, including optional ones. `create_parameter_fields` never supplies a default, so `get_historical_sector_pe` requires `from`, `to` and `exchange`. Pre-existing, but this PR takes the registry from 75 to 168 endpoints, so ~93 more tools now carry the over-required schema. Needs the `param.default` / `mandatory_params` semantics reconciled — not a drive-by.
- **#129** — `batch` (22/30), `sec` (9/12) and `transcripts` (2/4) carry 33 missing hints and are invisible to every guard here, because `_get_endpoint_groups()` does not include them. Registering them and filling the hints has to land together or CI goes red in between. `index` is clean.
- **#130** — `institutional.cik_mapper_by_name` is a functional duplicate of `cik_mapper`: identical path and params, with the name filtering living in the client wrapper that LangChain bypasses. Dropping the `name` hint here was correct against the endpoint, but it turned a visible validation failure into an invisible dead tool. Three options in the issue; wants a design decision.
- **#131** — `InstitutionalOwnershipDates.cik` now rejects an integer CIK in *every* validation mode, since `model_validate` runs before the mode branching. The change is still a net improvement and matches convention; whether FMP can emit a bare int needs a live probe, which I could not do without a key.

### Reviewer concerns checked and dismissed

- **Could a non-`ValueError` escape `register_batch` and abort the batch again?** No. Pydantic v2's `ValidationError` subclasses `ValueError`, so model-construction failures are caught by the batch loop.
- **Did the mypy pass weaken any test?** No. Every added `assert isinstance(...)` / `assert x is not None` is net-new; the `type: ignore[method-assign]`s sit only on mock assignments; `warn_unused_ignores` stays on for `tests.*`, so a dead suppression fails CI.

### Verification

`1090 passed, 3 skipped` · `ruff check` + `ruff format --check` clean · `mypy fmp_data tests` clean across 190 files.

### One thing that needs you

`dev` carries a `required_signatures` ruleset and there is no signing key on this machine, so `mergeStateStatus` stays `BLOCKED` however green CI is and `--auto` will arm but never fire. Auto-merge is enabled; the actual merge needs `gh pr merge 127 --squash --admin`.

MehdiZare:
## Second review round: findings and how each was handled

Three independent reviewers went over `ba8fd9b..1d3abfe` — one full-diff, one auditing runtime correctness by introspecting live objects, one attacking the guard tests by breaking each one to see which ones do not guard. Every claim below was re-verified before acting on it.

### The headline finding

**An endpoint could be deleted from a client's endpoint map with all 1090 tests green.** Every guard in `test_catalog_registration.py` iterates the endpoint map, so removing an entry just shrinks the universe they check — `setup_registry` silently went 168 → 167 and nothing failed. Deleting `get_profile` was caught only incidentally, by a `KeyError` in an unrelated fixture.

That is exactly the #121 regression class this file exists to catch. Fixed in `72a77f8` with an assertion that looks the other way — semantics → endpoints — against an explicit `CLIENT_SIDE_METHODS` allowlist. `company_logo_url` is the only legitimate orphan across all nine groups, measured.

### Fixed in this PR

| # | Finding | Commit |
|---|---|---|
| 1 | **Two intelligence tools advertised capabilities they do not have.** Repairing the 27 drifted endpoints made both register as live LC tools for the first time. `stock_news` said "stock-specific news" with `"Get stock news for AAPL"` as its lead example, but `STOCK_NEWS_ENDPOINT` takes only `page`/`start_date`/`end_date`/`limit` — and LC's `create_tool` calls `client.request()` directly rather than the sync method that delegates on `symbol`, so symbol-scoped queries matched a tool that ignored the symbol. It is in `DEFAULT_TOOLS`, so this was user-facing on MCP too. `stock_news_sentiments` is deprecated and returns `[]` without calling upstream; `docs/mcp/tools.md` already said so, the semantics table was the one place that did not. | `52a6ac3` |
| 2 | **A whole client module could leave the LC registry unnoticed.** Discovery knows 13 clients, the registry enrolls 9, and no guard could see the difference. Pinned as an explicit `LC_EXCLUDED_CLIENTS` set so the exclusion is declared rather than incidental. | `7b9a760` |
| 3 | **Nothing asserted `DEFAULT_TOOLS ⊆ catalog`.** A phantom entry failed only the per-section counts, and every message blamed the doc — so bumping the doc count made CI green while leaving a manifest entry that fails at MCP server startup. | `72a77f8` |
| 4 | **The alias guard compared `parameter_hints` only,** so a wrong `SemanticCategory` on an alias-only entry passed everything. Inert today, but `client_name` is already checked for aliases, so the omission was inconsistent rather than deliberate. | `72a77f8` |
| 5 | **`register()` logged tolerated skips at ERROR with a traceback.** pydantic's `ValidationError` subclasses `ValueError`, so a malformed pair emitted an error-level traceback and was *then* silently absorbed into `failures`. Split: `ValueError` logs at DEBUG (the caller reports it), anything else keeps ERROR because it genuinely aborts the batch. | `e41141f` |
| 6 | **`NAME_HINT` was left dead by this branch** — this PR removed its only consumer. Ruff does not flag unused module-level constants. | `0030af1` |
| 7 | Three `register_batch` behaviours the contract claims but nothing tested: empty batch is a no-op; a structurally malformed pair is a skip not a crash, and the rest of the batch still registers; a non-`ValueError` propagates. | `72a77f8` |
| 8 | `test_every_endpoint_passes_registry_validation` counted every endpoint-map key, so a lost semantics entry would surface as a bogus "name collision". The vector-store fixture discarded `register_batch`'s return. A blanket `type: ignore[call-arg]` hid every argument error on a model constructor. | `7b9a760`, `d627eb4` |
| 9 | **Two CHANGELOG claims were false.** "`EndpointRegistry` is not exported from `fmp_data.lc`" — it resolves today, it is merely absent from `__all__`. And the fail-loud example was unreachable: `get_endpoint_groups()` at the top of `setup_registry()` was never inside the removed handler. | `b5431ea` |

Every new guard was verified non-vacuous by introducing the drift it targets and confirming the specific failure — including the two added this round (dropping `technical` from the group map, and corrupting an alias's category).

### Filed as issues rather than widening this PR

- **#133** — `create_vector_store` wraps `setup_registry` in a blanket `except Exception` and returns `None`, so the fail-loud half of the #121 fix never reaches a caller. Narrowing that catch changes the public error contract of an exported function, which needs a deliberate decision.
- **#134** — `_get_type_pattern` builds `^(A|B))$`, an uncompilable regex, for `string` params with `valid_values`. Reachable: 22 params across 15 endpoints (`period`, `interval`, `timeframe`, `name`). Pre-existing; fixing it flips those patterns from "raises on use" to "actually validates", which can newly reject values.
- **#135** — hint duplication widened rather than consolidated: `CIK_HINT` now exists in four modules with different extraction patterns.
- **#136** — three company methods are double-listed under singular and plural tool keys. Same question as #126 and #130.

### Reviewer claims checked and refuted

- **"`register_batch` swallows pydantic `ValidationError`, blurring the contract."** Correct on the mechanism, but this is the intended behaviour, not a defect — it is what makes "never raises on validation" cover structurally malformed pairs. Pinned with a test instead of changed.
- **A `TypeError` from a malformed pair aborts the batch.** It does not: pydantic converts it to `ValidationError`, which is a `ValueError`. The genuinely-escaping path is an `ImportError` during lazy validation-rule construction; that is what the new test pins.

### Verification

`1096 passed, 3 skipped` · `ruff check` + `ruff format --check` clean · `mypy fmp_data tests` clean across 190 files · package builds (sdist + wheel).

Both the counts this PR claims were independently re-derived by two reviewers replaying the old abort-on-first-failure path against an `origin/dev` tree: **75 → 168 registrations and 27 invalid endpoints, both exact**, as are the 224 catalog / 159 default doc totals.

On the question of whether type-checking `tests/` weakened anything: **no**. All 13 removed `assert` lines have equal-or-stronger replacements, collected test count is identical base vs head for the modified files, nine edits *added* assertions, and all 18 added `# type: ignore`s were verified to be mock/monkeypatch artifacts rather than hidden defects. `warn_unused_ignores` was confirmed live for `tests.*` by injecting dead ignores and watching mypy fail.

The new guards were also confirmed to run in a core-only environment (no `lc`/`mcp` extras): 13 passed, 0 skipped. A guard that skips in CI is not a guard.

MehdiZare:
### Round closed — last residual fixed

The test reviewer re-verified every fix at `72a77f8` and re-broke each one to confirm it fails for the right reason. Its non-vacuity table is now 18/18 caught, including the four scenarios that slipped through at `1d3abfe`:

| Drift | At `1d3abfe` | Now |
|---|---|---|
| Drop `get_intraday_prices` from `COMPANY_ENDPOINT_MAP` | **not caught** — 1090 green | caught, names both the endpoint and the alias |
| Drop `get_profile` from `COMPANY_ENDPOINT_MAP` | incidental `KeyError` in an unrelated fixture | caught by design |
| Wrong `SemanticCategory` on alias-only `intraday_price` | **not caught** — 13 green | caught, names actual vs expected |
| Phantom entry in `DEFAULT_TOOLS` | caught, but every message blamed the doc | caught, names the phantom |

Its one remaining item — `cast(dict[str, Any], ...)` in `tests/unit/lc/test_mapping.py` widening element types — is fixed in `fe6608b` rather than deferred. `ENDPOINT_GROUPS` mixes the maps with a `display_name` string, so its value type genuinely is heterogeneous and a cast cannot be removed; it now names the real element types (`Endpoint[Any]`, `EndpointSemantics`) instead of `Any`, so the contents stay type-checked.

**Nothing from any of the three reviewers is left open.** Seven findings fixed here as isolated commits, four filed as #133–#136.

**Verification on a settled tree at `d627eb4`** (re-run after all edits were committed, since an intermediate reviewer run had raced a mid-edit working tree): `1096 passed, 3 skipped` · `mypy fmp_data tests` clean across 190 files · `ruff check` + `ruff format --check` clean · sdist + wheel build. Core-only environment (no `lc`/`mcp` extras): 992 passed, 5 skipped, with all 19 guard tests running rather than skipping.

### Release note

This PR targets `dev`, so merging it does not cut a release — `release.yml` fires on a PR merged to `main` carrying a `release:*` label. The CHANGELOG is therefore left as `## Unreleased` here on purpose: the version number is derived from that label at merge-to-`main` time, so cutting `## [2.5.0]` now would be guessing. Same pattern as #118, which cut 2.4.0 in its own release PR.

One thing worth fixing before the release PR: `dev` is missing the two commits on `main` that include the v2.4.0 release commit, so `git describe` from `dev` returns `v2.3.0` instead of `v2.4.0`. The real release is unaffected (`release.yml` runs on the merge commit into `main`, where the tag is reachable), but `dev-release.yml` derives its TestPyPI version from that tag, so dev builds are currently labelled against 2.3.0. Merging `main` into `dev` clears it.

PR #138, opened 2026-08-08
fix: uncompilable enum regex and integer CIK rejection (#134, #131)

### LINKED ISSUE #131: InstitutionalOwnershipDates.cik rejects an integer CIK regardless of validation_mode
## What

#127 made `cik` a declared field on `InstitutionalOwnershipDates` (`fmp_data/institutional/models.py`) instead of letting it fall through to `extra="allow"`:

```python
cik: str | None = Field(default=None)
```

`BaseClient._validate_model` calls `model.model_validate(payload)` **before** any `validation_mode` branching, so a pydantic type error raises whatever the mode is set to:

```python
>>> InstitutionalOwnershipDates.model_validate(
...     {"date": "2024-03-31", "year": 2024, "quarter": 1, "cik": 320193}
... )
ValidationError: cik — Input should be a valid string
```

A string CIK parses fine. Previously an integer landed in `model_extra` and parsed under `lenient`/`warn`.

## Assessment

The change is a net improvement — it removes an "unexpected fields" hard failure under `strict` mode, and `str | None` matches the convention used for `cik` in `sec/models.py`, `investment/models.py` and `intelligence/models.py`. This is not a reason to revert it.

The open question is factual and **unverified**: does FMP's `institutional-ownership/dates` ever return `cik` as a bare integer? CIKs are zero-padded identifiers, so a string is the sensible representation, but JSON producers drop leading zeros routinely. No API key was available during review, and VCR cassettes are gitignored so they are not checkable from a clean checkout.

## Suggested resolution

1. Probe the live endpoint and record what type `cik` actually comes back as.
2. If it can be an integer, add coercion rather than widening the type — a `field_validator(mode="before")` that `str()`s an int, or `model_config = ConfigDict(coerce_numbers_to_str=True)` — so the field stays `str | None` and callers keep one type.
3. Apply the same decision to the other `cik` fields across the codebase, which have the same exposure.

Found during review of #127.

### LINKED ISSUE #134: EndpointBasedRule._get_type_pattern builds an uncompilable regex for enum string params
## Problem

`fmp_data/lc/registry.py:272` has an unbalanced parenthesis:

```python
case "string":
    if valid_values:
        return [f"^({'|'.join(map(str, valid_values))}))$"]
```

This produces `^(annual|quarter))$` — one closing paren too many:

```python
>>> re.compile("^(a|b))$")
re.error: unbalanced parenthesis at position 6
```

## Reachability

Not dead code. `_get_type_pattern` is called by `EndpointBasedRule.get_parameter_requirements` (`registry.py:299`), which is called by `ValidationRuleRegistry.get_parameter_requirements` (`registry.py:370`). Patterns from that family are consumed by an uncaught `re.match` in `fmp_data/lc/validation.py:140`.

**22 parameters across 15 endpoints** hit the `string` + `valid_values` branch, measured against the live endpoint definitions:

- `period` — `fundamental.get_income_statement`, `get_balance_sheet`, `get_cash_flow`, `get_key_metrics`, `get_financial_ratios`, `get_full_financial_statement`; `company.get_analyst_estimates`, `get_product_revenue_segmentation`, `get_geographic_revenue_segmentation`
- `interval` — `alternative.get_crypto_intraday`, `get_forex_intraday`, `get_commodity_intraday`
- `timeframe` — `technical.get_sma`, `get_ema`
- `name` — `economics.get_economic_indicators`

Practical impact today is limited because no in-repo caller reaches `ValidationRuleRegistry.get_parameter_requirements` on a live registry, so the bad pattern is built but never compiled in normal use. It is a live landmine on a public method.

## Fix

Drop the extra paren:

```python
return [f"^({'|'.join(map(str, valid_values))})$"]
```

Add a test that compiles every pattern returned by `get_parameter_requirements` for all registered endpoints, so this cannot regress.

## Why not fixed in #127

Pre-existing and untouched by that PR. Fixing it flips these patterns from "always raises on use" to "actually validates", which could newly reject values that currently slip through — a behavior change that deserves its own PR and its own test.

Note `economics.get_economic_indicators` passes enum members rather than strings, so `map(str, ...)` yields `EconomicIndicatorType.GDP` rather than `GDP`. That should be fixed in the same pass — use `.value` where present.

Found during review of #127.

MehdiZare:
Closes #134
Closes #131

Spec: `docs/superpowers/specs/2026-08-07-open-issues-remediation-design.md` §4 PR1.

## #134 — uncompilable enum regex

`_get_type_pattern` emitted `^(annual|quarter))$`. Reachable via
`EndpointBasedRule.get_parameter_requirements` ->
`ValidationRuleRegistry.get_parameter_requirements`, consumed by an
uncaught `re.match` in `validation.py:140`. 22 parameters across 15
endpoints.

Also unwraps enum members to `.value` — `economics.get_economic_indicators`
would otherwise compile a pattern that rejects every valid value — and
escapes alternatives, since `valid_values` is endpoint-declared data.

**Reviewers: this is a behaviour change.** These patterns previously
raised `re.error` on any use and now validate correctly. No in-repo caller
reaches them today, so library users on the default paths see no change;
an external caller invoking the public
`ValidationRuleRegistry.get_parameter_requirements` now gets working
validation instead of a crash. Matching is case-sensitive against the
endpoint's declared `valid_values` — `period="annual"` is accepted,
`period="ANNUAL"` is not. That is the intended fix and the reason #134
was kept out of #127.

## #131 — integer CIK

Live probe (2026-08-07) found `institutional-ownership/dates` returns no
`cik` field at all — only `date`, `year`, `quarter` — so the
`ValidationError` in the issue is unreachable from that endpoint. Where
`cik` *is* returned (`institutional-ownership/latest`,
`sec-filings-search/cik`) it is consistently a 10-digit zero-padded string.

New `fmp_data.models.CIK` type coerces int -> `"%010d"` and leaves strings
alone; applied to all 58 CIK-valued fields across 8 modules — 45 named
`cik` plus 13 named `*_cik` (`reporting_cik`, `company_cik`,
`targeted_cik`, `intermediary_commission_cik`). Padding rather than bare
`str()` so a caller joining on `cik` across two endpoints gets one
spelling.

## Tests
- `test_every_emitted_pattern_compiles` — compiles every pattern the
  registry can emit across all 168 registered endpoints
- `test_enum_valid_values_use_wire_value_not_repr`
- `TestCIKCoercion` — int padding, string pass-through, None, required-ness, bool rejection
- `test_no_model_declares_a_bare_str_cik` — guards against a new model
  reintroducing the bug class
- Guard inspects 61 CIK-valued fields across 54 classes with a floor
  assertion, so it cannot pass vacuously if an extras-gated subtree fails
  to import

---

## Follow-up review (post-review pass)

A review pass over this branch found three items; all are addressed here as
isolated commits. Full detail in the summary comment below.

### `09dca16` — enum unwrapping moved to the source
`_get_type_pattern` passed `value.value` straight to `re.escape`, which
raises `TypeError` for any `Enum` whose value is not a `str` — trading a
wrong-pattern bug for a crash bug in the function this PR exists to harden.
`EndpointParam.__post_init__` now unwraps enum members once, so both
consumers of `valid_values` see the wire value. Only `Enum` is unwrapped:
`validate_value` compares against a *converted* value, so stringifying
`transcripts.quarter`'s `[1, 2, 3, 4]` would break every such comparison.

### `135309f` — request path (scope expansion)
#131's fix covered responses only. Outbound, every `cik` param was
`ParamType.STRING`, so `cik=320193` was sent as `"320193"` — a request that
succeeds and returns nothing, worse than the inbound rejection it mirrored.
New `ParamType.CIK` applied to all 16 declarations across 6 modules; client
signatures widened to `str | int`; the LangChain arg models use `CIK`.
It pads numeric strings too, which the response coercer deliberately does
not — inbound, re-padding misreports what the API sent; outbound the padded
form is simply the correct request.

### `a7c3108` — drift guard widened
The guard claimed to cover every CIK-valued field but walked only
`*.models`. The gap sat exactly where drift had already happened: the three
`*Args` models in `*.schema` were all bare `str` and invisible to it.

## Filed, not fixed here
- **#139** — `fmp_data.investment.schema` is un-importable under pydantic
  2.13 (`ETFHoldingsArgs` declares `date: date`). Pre-existing on `dev`,
  surfaced by widening the guard's walk. Fixing it means renaming a public
  field on a LangChain-facing arg model, well outside this PR. Carried as an
  explicit allowlist entry rather than a silent skip, so coverage resumes
  automatically once it is fixed and a *newly* broken module fails loudly.

Suite: 1118 passed, 3 skipped (from 1105). ruff + mypy clean.
Every new guard verified non-vacuous by mutation.


MehdiZare:
## Review findings — disposition

No inline review threads exist on this PR to reply to: CodeRabbit skipped it (auto-review is disabled for non-default base branches, and this targets `dev`), and Codecov left only a status comment. The findings below come from a manual review pass over the branch diff, recorded here so the reasoning is on the record rather than only in commit messages.

| # | Finding | Disposition |
|---|---------|-------------|
| F1 | `_get_type_pattern` crashes on non-`str` enum values | Fixed — `09dca16` |
| F2 | CIK drift guard claimed more coverage than it checked | Fixed — `a7c3108` |
| F3 | Request path still rejects/mismatches integer CIKs | Fixed — `135309f` |
| F4 | `_coerce_cik(-320193)` → `"-000320193"` | No action |
| — | `fmp_data.investment.schema` un-importable | Filed as #139 |

### F1 — enum stringification crashes (`fmp_data/lc/registry.py`)

The new enum handling passed `value.value` straight to `re.escape`, which rejects a non-`str`:

```pycon
>>> EndpointBasedRule._get_type_pattern("string", list(SomeIntEnum))
TypeError: decoding to str: need a bytes-like object, int found
```

The pre-#134 code (`str(value)`) produced a *wrong* pattern for that input but never raised, so as written this PR traded a wrong-pattern bug for a crash bug inside the very function it exists to harden. Unreachable today — `EconomicIndicatorType` is a `str, Enum`, and the only integer `valid_values` in the repo sits on an `INTEGER`-typed param that never consults them — but latent.

**Fixed at the source rather than the symptom.** `EndpointParam.__post_init__` unwraps enum members once, so neither consumer of `valid_values` special-cases `Enum`. Only `Enum` is unwrapped and values keep their native type: `validate_value` compares the *converted* request value against this list, so stringifying `transcripts.quarter`’s `[1, 2, 3, 4]` would make every membership check fail. `_get_type_pattern` keeps a defensive `str(value.value)` for direct callers of the staticmethod.

### F2 — guard scope did not match its claim (`tests/unit/test_models.py`)

The docstring said *“every CIK-valued field must route through the CIK coercer”*, but the walk filtered to modules named `*.models`. Request-argument models live in `*.schema` — and all three of them (`Form13FArgs`, `Form13FDatesArgs`, `PortfolioDateArgs`) were bare `str`, i.e. the guard was blind in precisely the place drift had already occurred.

Now walks every module. Fields are keyed on their defining module so the floor counts distinct fields rather than repeat sightings via `dir()`.

### F3 — request path (scope expansion)

#131 was filed and fixed against the response path. The request path had the same defect with a nastier failure mode: every `cik` param was `ParamType.STRING`, so `cik=320193` went out as `"320193"` — FMP matches the zero-padded form, so the request *succeeds and returns nothing*. Silent empty results are worse than the inbound `ValidationError` this PR removes.

Rather than thread coercion through 26 client signatures, CIK became a first-class parameter type — `ParamType.CIK`, applied to all 16 declarations across 6 modules. One deliberate asymmetry, documented in the code:

> Inbound, re-padding a string would misreport what the API actually sent. Outbound there is nothing to misreport — the padded form is simply the correct request.

So the request converter pads numeric strings; the response coercer does not. Non-numeric strings pass through untouched so a bad value surfaces as an API error instead of being mangled into one, and `bool` is rejected to match pydantic’s response-side behaviour.

### F4 — no action

`_coerce_cik(-320193)` yields `"-000320193"`. A negative CIK is not a real input and the documented contract is best-effort coercion; adding a rejection path buys nothing and costs a branch.

### #139 — filed, not fixed

Widening F2’s walk surfaced that `fmp_data.investment.schema` cannot be imported **at all** under pydantic 2.13 — `ETFHoldingsArgs` declares `date: date`, clashing with the `datetime.date` import. Verified pre-existing on `dev`, so not a regression from this PR. Every arg model in that module is unreachable at runtime; it went unnoticed because nothing in the unit suite imports it.

Not fixed here: the repair means renaming a public field on a LangChain-facing arg model, which is a behaviour change well outside a PR scoped to CIK coercion and enum regex generation. It is carried as an **explicit allowlist entry**, not a silent skip — so a *newly* un-importable module fails the guard loudly, and coverage of this one resumes automatically once #139 lands.

### Verification

- Suite 1118 passed / 3 skipped, up from 1105; `ruff` and `mypy` clean
- Every new guard mutation-tested: reverting the `str()` reproduces the `TypeError`; neutering `__post_init__` fails the unwrapping assertion; flipping one `cik` param back to `STRING` fails the param guard; reverting `Form13FDatesArgs.cik` to bare `str` fails the widened field guard (and did **not** fail the old one); emptying the allowlist fails the un-importable-module assertion

PR #140, opened 2026-08-08
fix(investment): make schema module importable under pydantic 2.13 (#139)

### LINKED ISSUE #139: fmp_data.investment.schema is un-importable under pydantic 2.13 (field/type name clash)
Found while widening the CIK drift guard in #138. **Pre-existing on `dev`; not introduced by that PR.**

## Symptom

The module cannot be imported at all:

```console
$ python -c "import fmp_data.investment.schema"
pydantic.errors.PydanticUserError: Error when building FieldInfo from annotated attribute.
Make sure you do not have any field name clashing with a type annotation.
https://errors.pydantic.dev/2.13/u/unevaluable-type-annotation
```

## Cause

`fmp_data/investment/schema.py` imports `from datetime import date`, then `ETFHoldingsArgs` declares a field named `date` annotated with that same name:

```python
class ETFHoldingsArgs(SymbolArg):
    date: date = Field(
        description="Holdings date", json_schema_extra={"examples": ["2024-01-15"]}
    )
```

Under pydantic 2.13 the field name shadows the type before the annotation is evaluated, so class construction raises.

## Impact

Every arg model in the module is unreachable at runtime — `ETFHoldingsArgs`, `ETFInfoArgs`, `PortfolioDateArgs`, and the rest. Any LangChain tool registration path that imports it fails outright. It went unnoticed because nothing in the unit suite imports the module.

Worth checking whether other `*/schema.py` modules have the same clash (`date`, `datetime`, `type`, ...) and whether this is a regression from a pydantic upgrade or has always been broken.

## Why not fixed in #138

The fix means either renaming a public field on a LangChain-facing arg model or restructuring the datetime import — a behaviour/API change well outside that PR, which is scoped to CIK coercion and enum regex generation. #138 declares `PortfolioDateArgs.cik` with the new `CIK` type so it is correct once this is resolved, and its drift guard carries an explicit allowlist entry for this module rather than silently skipping it, so coverage resumes automatically when this is fixed.

## Suggested fix

Rename the annotation rather than the field, keeping the wire name stable:

```python
from datetime import date as date_type

class ETFHoldingsArgs(SymbolArg):
    date: date_type = Field(...)
```

Then drop the allowlist entry in `tests/unit/test_models.py::TestCIKCoercion::test_no_model_declares_a_bare_str_cik`, and add an import-smoke test that every `fmp_data.*` module imports cleanly — that is the guard whose absence let this sit.

MehdiZare:
Closes #139

## The break

`fmp_data.investment.schema` could not be imported at all:

```console
$ python -c "import fmp_data.investment.schema"
pydantic.errors.PydanticUserError: Error when building FieldInfo from annotated attribute.
Make sure you don't have any field name clashing with a type annotation.
```

`ETFHoldingsArgs` and `MutualFundHoldingsArgs` each declared a field named `date` annotated with the `date` imported from `datetime`. Under pydantic 2.13 the field name shadows the type before the annotation is evaluated, so class construction raises and every arg model in the module is unreachable at runtime. Any LangChain registration path importing it failed outright.

Pre-existing on `dev`; surfaced while widening the CIK drift guard in #138.

## The fix

Import the annotation as `dt_date`, matching the convention already used in `fmp_data/technical/schema.py`. The **field is still named `date`**, so the wire format is unchanged:

```python
from datetime import date as dt_date

class ETFHoldingsArgs(SymbolArg):
    date: dt_date = Field(...)
```

Verified the field name, annotation and parsing are all preserved:
```
ETFHoldingsArgs field=date annotation=<class 'datetime.date'> required=True
parse: 2024-01-15
```

## The guard that was missing

The one-line fix matters less than why nothing caught this. **No test imported the module.** New `tests/unit/test_imports.py` walks every module in the package and asserts it imports.

It fails on *any* exception, not just `ImportError` — #139 raised `PydanticUserError`, so an `ImportError`-only check would have sailed past it. Subpackages behind optional extras (`fmp_data.lc`, `fmp_data.mcp`) are allowed to raise `ImportError`/`SystemExit` when the extra is absent, since that is their documented behaviour; anything else, anywhere, fails the test. It carries a floor assertion so it cannot pass vacuously if the walk stops yielding.

Confirmed red before the fix, green after.

## Knock-on

The `_KNOWN_UNIMPORTABLE = {"fmp_data.investment.schema"}` allowlist that #138 had to add is now empty. The CIK drift guard covers the module again — coverage rises from 63 to 64 CIK-valued fields, and it confirms `PortfolioDateArgs.cik` was already correct.

## Verification

- `make lint` clean
- Full unit suite: 1126 passed, 2 skipped
- Only 1 module was un-importable package-wide; a scan of every `*/schema.py` for the same field/type clash found no others

## Note for reviewers

`Closes #139` does not fire on dev-base PRs in this repo — the issue needs closing by hand after merge.


MehdiZare:
## Code review round — all findings addressed

Two independent reviewers went over `5e7556d..75fe95d`: one on architecture/production correctness, one adversarial on the new guard. Findings were verified by **mutation testing**, not inference. Everything actionable is now fixed in `f7e03bc` + `429b9b6`, or filed as a follow-up.

### What held up

The production fix itself needed no change. Independently verified:

- `dt_date` is the right alias. `from __future__ import annotations` was tested and does **not** fix the clash (pydantic still resolves `date` through the class namespace); renaming the field would be a wire-visible break; `import datetime` diverges from repo convention.
- Wire format unchanged: `model_fields` → `['symbol', 'date']`, `model_dump()` → `{'symbol': 'SPY', 'date': '2024-01-15'}`, JSON schema `format: date` — identical before and after.
- The guard is genuinely RED at `5e7556d` with the exact `PydanticUserError` from #139, GREEN at HEAD. The broad `except Exception` is load-bearing.
- `mypy` was clean on the *broken* pre-fix file — static tooling cannot catch this class, which is what justifies a runtime guard.

### Fixed — the guard was under-delivering on its coverage claim

Each of these was proven by injecting a real fault, not argued:

**1. The optional-extra exemption was keyed on module *name*, so it applied even when the extra was installed.** Appending a genuine bad import to `fmp_data/lc/vector_store.py` and `fmp_data/mcp/server.py` with extras present left the suite **green** — in exactly the `langchain` / `mcp-server` CI sessions meant to catch it. 20 of 126 modules were effectively unguarded.
Now keyed on `importlib.util.find_spec`, so the exemption disappears whenever the extra is actually installed. Same mutation now fails loudly. Also asserts the exempted set cannot silently grow past the four modules that genuinely need an extra.

**2. `walk_packages` ran without `onerror`, silently dropping whole subtrees.** Breaking `fmp_data/lc/__init__.py` took the walk from 126 modules to 117 without a word — the test only failed via collateral damage from unrelated importers. Now has an `onerror` handler that asserts.

**3. The floor of `> 50` sat at 40% of the true count (126).** 60% of the package could vanish unnoticed. Raised to 120.

**4. Nothing tested the fix's actual behavior.** No test constructed `ETFHoldingsArgs` or `MutualFundHoldingsArgs`, so renaming the field to `holding_date` to dodge the shadowing — a silent wire-format break — would have passed everything while `test_imports.py` stayed green. Now pinned in `tests/unit/test_investment.py`.

**5. `find_spec` can raise rather than return `None`** (a `meta_path` finder may do this). Now treated as absent instead of erroring.

### Corrected claims

- **"61 to 64 CIK fields" was wrong — it is 63 to 64.** Ran `_cik_fields_via_import()` at both commits: base 63, HEAD 64. The 61 was carried over from #138's starting state; this PR's actual contribution is the single `PortfolioDateArgs.cik`. Fixed in the CHANGELOG and this PR body.
- **"Any LangChain registration path importing it failed outright" overstated the blast radius.** Nothing in `fmp_data/` imports `investment.schema` at all. Reworded — the orphan status is precisely *why* the break survived.

### Verification

- Real no-extras venv (the CI `core` session): 126 modules walked, 0 broken, 0 walk errors, exactly the 4 expected exemptions
- Full suite: **1126 passed, 4 skipped**
- `ruff check`, `ruff format --check`, `mypy` clean

### Deliberately not fixed here

One assertion was **removed** rather than added: the new test does not assert `is_required()` on the `date` field. `ETF_HOLDINGS` declares that parameter optional while `MUTUAL_FUND_HOLDINGS` declares it mandatory, and both arg models declare it required — a three-way contradiction. Pinning either side would cement half of it and make reconciliation harder. Tracked in #143, with a comment in the test pointing there.

### Follow-ups filed

- **#141** — `investment`, `institutional` and `intelligence` ship `schema.py` modules that **nothing in the package imports** (31 arg-model classes). This is *how #139 hid for so long*. Either wire them in like the six domains that do, or delete them.
- **#142** — 23 latent `date`-field/type shadowing sites across 7 files. All safe today (verified: only the exact `date: date` clash raises), but each is one annotation edit from reproducing #139. Suggested an AST guard so the trap cannot be re-introduced.
- **#143** — `etf_holdings` vs `mutual_fund_holdings` disagree on the `date` parameter's optionality *and* `ParamType` (`DATE` vs `STRING` — same drift class as #131/#134).

All three are production/contract changes that need a deliberate decision, so they are out of scope for a one-line import fix.

Issue #144, opened 2026-08-08
EndpointParam.required is dead metadata for 13 optional params


MehdiZare:
Found during review of the #128/#129 PR.

## What

13 parameters live in an endpoint's `optional_params` while declaring `required=True`. The two statements contradict each other, and the `required` flag is the one that is ignored.

Examples: `alternative.get_crypto_historical.start_date` / `end_date`, and several of the intelligence calendar endpoints.

## Why it is inert

`Endpoint.validate_params` skips a `None` value for any param not in `mandatory_params` **before** `validate_value` can consult `required`. So a caller omitting these succeeds:

```python
CRYPTO_HISTORICAL.validate_params({"symbol": "BTCUSD"})
# -> {'symbol': 'BTCUSD'}   no error, despite required=True
```

The #128 fix (LangChain tool schemas) correctly keys off `mandatory_params` membership rather than the `required` flag, so generated tools are right. But nothing flags the contradiction, so the next reader may reasonably trust `required=True` and conclude the param is mandatory when it is not.

The mirror case also exists and is *not* a defect: 13 params in `mandatory_params` carry a non-None `default`. Those are genuinely required — `SENATE_LATEST.validate_params({})` raises `Missing mandatory parameter: page` — so the default is a suggested value, not an exemption. Worth documenting rather than changing.

## Options

1. Drop `required=True` from the 13 optional params, so the flag stops contradicting the list the param sits in.
2. Move them into `mandatory_params`, if any genuinely are required — needs checking against the live API per endpoint, since that changes caller-visible behaviour.
3. Remove `EndpointParam.required` entirely and derive requiredness from list membership, which is the only thing any consumer actually reads.

Option 3 is the honest one — a field that no code path consults is worse than no field, because it reads as authoritative. It is also the largest change and touches every endpoint definition.

## Guard worth adding either way

A test asserting that no param in `optional_params` declares `required=True` (and vice versa), so the two representations cannot drift apart again.

## Scope

Pre-existing; not introduced by #128/#129. No runtime bug today — generated tool schemas and request validation are both correct. This is about a lie in the data model that will mislead someone eventually.

MehdiZare:
Additional context from the #145 review, which is where this contradiction now has user-visible consequences.

#145 makes `create_parameter_fields` take mandatory and optional params as separate lists, so requiredness in the generated tool schema comes from list membership alone. The 13 params that sit in `optional_params` while declaring `required=True` therefore became **omittable by an LLM**:

[...]

Issue #136, opened 2026-08-07
MCP catalog double-lists three company methods under singular and plural keys


MehdiZare:
## Problem

`discover_all_tools()` returns 224 entries for 168 endpoints. Beyond the ambiguity tracked in #126 and the duplicate tracked in #130, three `(client, method)` pairs are advertised twice under different tool keys:

| Method | Keys |
|---|---|
| `company.get_historical_prices` | `historical_price`, `historical_prices` |
| `company.get_intraday_prices` | `intraday_price`, `intraday_prices` |
| `company.get_executives` | `executives`, `key_executives` |

Two names for one callable means an MCP client sees two tools that do exactly the same thing, and an LLM has to pick between them with nothing to distinguish them.

## Context

#127 deliberately *synced* the singular aliases rather than removing them — `fmp_data/company/mapping.py:957` and `:1028` — because the `intraday_price` alias had drifted and was missing three parameter hints. Syncing was the right call for that PR: removing a public tool key is not a fix, it is an API change.

The duplication itself remains, and now both halves are correct and identical, which arguably makes it more confusing rather than less.

## Options

1. Keep one key per method, deprecate the other over a release (announce in CHANGELOG, keep resolving for one minor version).
2. Keep both but exclude the aliases from `DEFAULT_TOOLS` so the default server surfaces one of each.
3. Declare the aliases intentional and document why, so future guards stop flagging them.

Worth settling together with #126 and #130 — all four are the same question about what the tool-key namespace guarantees.

Found during review of #127.

MehdiZare:
Fixed by #147, merged to `dev` as `0d04519`.

Resolved via a combination of **option 1 and option 2**, which #147 treats as one answer to the shared question you raised at the bottom of this issue: *one tool key per `(client, method)` pair.*

| Deprecated key | Replacement |
|---|---|
| `company.executives` | `company.key_executives` |
| `company.historical_price` | `company.historical_prices` |
| `company.intraday_price` | `company.intraday_prices` |

- Deprecated keys still resolve through 2.6 in both bare and fully qualified form, emitting a warning naming the replacement, and are removed at 3.0 (catalog 223 → 220).
- They leave `DEFAULT_TOOLS` now (159 → 156), so the default server already advertises exactly one tool per method.

Two things worth recording here.

[...]

Issue #135, opened 2026-08-07
Consolidate duplicated ParameterHint constants into fmp_data.lc.hints


MehdiZare:
## Problem

#127 added shared hints to `fmp_data/lc/hints.py` (`PAGE_HINT`, `SECTOR_HINT`, `INDUSTRY_HINT`, `YEAR_HINT`, `QUARTER_HINT`, `CIK_HINT`, `FROM_TO_DATE_HINTS`) but did not retire the per-module copies. The result is addition, not consolidation — the same concept now exists under the same name in several places with *different extraction patterns*:

- `CIK_HINT` — `fmp_data/lc/hints.py:47`, `fmp_data/company/hints.py:11`, `fmp_data/institutional/mapping.py:36`, `fmp_data/investment/mapping.py:62`
- `PAGE_HINT` — `fmp_data/lc/hints.py:58`, `fmp_data/institutional/mapping.py:56`, `fmp_data/fundamental/mapping.py:105`, `fmp_data/intelligence/mapping.py:181`
- `LIMIT_HINT` — `fmp_data/fundamental/mapping.py:87`, `fmp_data/intelligence/mapping.py:191`

`fmp_data/intelligence/mapping.py` now imports the shared `CIK_HINT` while `fmp_data/institutional/mapping.py` keeps its local copy, so one file's `cik` parameter extracts differently from another's for no principled reason.

## Why this matters

These hints drive natural-language parameter extraction for LangChain tools. Divergent `extraction_patterns` for the same parameter mean the same user phrasing resolves on one tool and not another, and the divergence is invisible — nothing compares them.

## Fix

Move each concept to `fmp_data/lc/hints.py` with one agreed pattern set, re-export or import from the domain modules, and delete the duplicates. Consider a test asserting no two modules define a `ParameterHint` under the same name with different contents.

## Why not fixed in #127

It touches every mapping module and changes extraction behavior wherever the winning pattern differs from the local one. That is a behavioral change across the whole LC surface and does not belong inside a registry-resilience fix.

Found during review of #127.

MehdiZare:
Fixed by #150 (merged 2026-08-08).

Ten `ParameterHint` names were bound in more than one module with different contents — `SYMBOL_HINT` in six, `CIK_HINT` and `PAGE_HINT` in four each. Each concept now has one definition in `fmp_data.lc.hints`, imported by the domain modules.

[...]

PR #145, opened 2026-08-08
fix(lc): optional params in tool schemas, and register the four unguarded clients (#128, #129)

### LINKED ISSUE #128: LangChain tools mark every endpoint parameter as required, including optional ones
## What

`ToolFactory.create_parameter_fields` (`fmp_data/lc/vector_store.py`) builds every parameter as `Field(description=...)` with **no default**, so pydantic treats it as required. `get_field_type(param.param_type, optional=(param.default is not None))` only widens the *type* to `| None`; it never supplies a default.

The result is that optional endpoint parameters become required arguments in the generated LangChain tool schema. An LLM calling the tool has to invent values for parameters the endpoint does not need.

## Reproduction

```python
from fmp_data.lc.vector_store import ToolFactory
from fmp_data.market.endpoints import HISTORICAL_SECTOR_PE as ep
from fmp_data.market.mapping import MARKET_ENDPOINTS_SEMANTICS
from pydantic import create_model

sem = MARKET_ENDPOINTS_SEMANTICS["historical_sector_pe"]
fields = ToolFactory.create_parameter_fields(
    list(ep.mandatory_params) + list(ep.optional_params or []), sem.parameter_hints
)
M = create_model("Args", **fields)

print([n for n, f in M.model_fields.items() if f.is_required()])
# ['sector', 'from', 'to', 'exchange']
print([p.name for p in ep.mandatory_params])
# ['sector']
```

`from`, `to` and `exchange` are optional on the endpoint but required in the tool schema.

## Why it matters now

This is pre-existing, not a regression. But #121 took the registry from 75 registered endpoints to 168, so roughly 93 additional endpoints now emit LangChain tools carrying this over-required schema. The blast radius grew by more than 2x.

## Suggested fix

Give optional parameters a pydantic default (`None`, or `param.default` when the endpoint declares one) so `is_required()` is false:

```python
default = param.default if param.default is not None else (... if is_mandatory else None)
param_fields[param.name] = (field_type, Field(default, description=description))
```

Needs a decision on how `param.default` interacts with mandatory-vs-optional — the current `optional=(param.default is not None)` heuristic is not the same question as "is this in `mandatory_params`", and those two should probably be reconciled at the same time.

## Test to add

A guard asserting that for every registered endpoint, the generated tool schema's required-field set equals `{p.name for p in endpoint.mandatory_params}`.

Found during review of #127.

### LINKED ISSUE #129: batch, sec and transcripts semantics are unguarded and carry 33 missing parameter hints
## What

`_get_endpoint_groups()` in `fmp_data/lc/registry.py` registers nine client groups (alternative, company, economics, fundamental, market, technical, institutional, intelligence, investment) — 168 endpoints. The `batch`, `index`, `sec` and `transcripts` modules are **not** in that dict.

Those four are still live MCP tools (they make up the 224-tool catalog vs the 168-endpoint registry), but because they never reach `EndpointRegistry.validate_endpoint()`, nothing checks their `parameter_hints` against their endpoints. 33 endpoints there declare `parameter_hints={}` while their endpoint takes real parameters.

The #127 guard tests cannot see any of this — they iterate `get_endpoint_groups()`.

## The drift

All 33 are **missing** hints. There are zero dead hints anywhere in the codebase.

| module | endpoints | drifted | missing hints |
|---|---|---|---|
| `batch` | 30 | 22 | `symbols`, `exchange`, `short`, `part`, `year`, `period`, `date` |
| `sec` | 12 | 9 | `cik`, `symbol`, `company`, `formType`, `from`, `to`, `page`, `limit`, `sicCode` |
| `transcripts` | 4 | 2 | `symbol`, `year`, `quarter`, `page`, `limit` |
| `index` | 6 | 0 | — clean |

Examples: `batch.batch_quote` (missing `symbols`), `batch.income_statement_bulk` (missing `period`, `year`), `sec.filings_search_form` (missing `formType`, `from`, `to`, `page`, `limit`), `transcripts.transcript` (missing `symbol`, `year`, `quarter`, `limit`).

## Impact

Parameter hints feed embedding text and the argument descriptions on generated tools. Missing hints mean an LLM sees a parameter with no natural-language description — degraded tool selection and argument filling, not a crash.

## Suggested fix

1. Add `batch`, `index`, `sec` and `transcripts` to `_get_endpoint_groups()`. This raises the guarded catalog from 168 to 224 and makes all 33 fail CI immediately.
2. Fill in the missing hints, reusing the shared constants in `fmp_data.lc.hints` (`PAGE_HINT`, `LIMIT_HINT`, `YEAR_HINT`, `QUARTER_HINT`, `CIK_HINT`, `FROM_TO_DATE_HINTS`, `SYMBOL_HINT`).

Step 1 before step 2 will red the build, so they want to land together — which is why this is not folded into #127.

Each of the four modules also needs a `SemanticCategory` decision, since `_get_endpoint_groups()` assigns one per group.

Found during review of #127.

MehdiZare:
Closes #128
Closes #129

## #128 — every tool parameter was marked required

`ToolFactory.create_parameter_fields` built each parameter as `Field(description=...)` with no default, so pydantic treated optional endpoint params as required. `get_field_type(..., optional=(param.default is not None))` widened the *type* to `| None` but never supplied a default. An LLM calling the tool had to invent values for parameters the endpoint does not need.

`HISTORICAL_SECTOR_PE` before / after:
```
required in schema: ['sector', 'from', 'to', 'exchange']   ->   ['sector']
endpoint mandatory: ['sector']
```

The issue flagged an unresolved question: `param.default is not None` is not the same question as "is this in `mandatory_params`". Those are now reconciled — the signature takes mandatory and optional params separately rather than inferring requiredness from the default. Both awkward cases were checked at the wire layer rather than reasoned about:

- **13 mandatory params carry a non-None default** (`intelligence.get_senate_latest.page=0`, …). They stay required with no schema default: `SENATE_LATEST.validate_params({})` raises `Missing mandatory parameter: page`, so giving them a schema default would emit tool calls the client rejects at runtime.
- **13 optional params declare `required=True`** (`alternative.get_crypto_historical.start_date`, …). They become optional: `validate_params` skips `None` for non-mandatory params before `validate_value` consults `required`, so omitting them already worked. That contradiction is filed separately as #144.

## #129 — four clients were unguarded, 33 endpoints had missing hints

`batch`, `index`, `sec` and `transcripts` were absent from `_get_endpoint_groups()`, so their semantics never reached `validate_endpoint()` and nothing checked their `parameter_hints`. Drift measured before the fix: batch 22, sec 9, transcripts 2, index 0 — **33 endpoints**.

Registry coverage: **168 -> 216 endpoints**, 9 -> 18 groups. (216, not 224: that figure is the MCP *tool* count. All 13 clients total 220 endpoints, minus 4 excluded collisions.)

### Groups are partitioned by category, not flattened

`batch` declares five different `SemanticCategory` values across its endpoints and `sec` declares two, but the registry model requires one category per group. Flattening them to a single per-module category would have been lossy in the one place it matters — `category` drives the semantic search the registry exists to serve, so relabelling `batch.income_statement_bulk` from Fundamental Analysis to Market Data stops fundamentals queries matching it.

Instead the groups are partitioned by the category each semantics actually declares, derived programmatically rather than hand-listed: `batch_market`, `batch_fundamental`, `batch_intelligence`, `batch_company`, `batch_institutional`, `sec_company`, `sec_intelligence`, plus `index` and `transcripts`. **Not one declared category changed.** A test asserts the partition helper is a no-op on the 11 already-uniform clients, proving it agrees with the hand-maintained structure.

### Four genuine collisions are excluded, as a stopgap

`get_crypto_quotes` and `get_forex_quotes` (alternative vs batch), `search_by_cik` (market vs sec), `get_profile` (company vs sec) are distinct methods on distinct clients. LangChain tool names come from `method_name`, so they cannot coexist — left alone, `sec`'s would silently overwrite `company`'s. They are excluded via a documented, test-pinned constant that cites #126 and #136; the exclusion list should shrink when those land. Endpoints excluded from LC enrollment: **52 -> 4**.

### Hints

60 of the 72 hint references reuse the shared constants in `fmp_data/lc/hints.py`. Five new local hints exist only where no shared concept applies (`SHORT_HINT`, `PART_HINT`, `FORM_TYPE_HINT`, `SIC_CODE_HINT`, `COMPANY_NAME_HINT`). `SYMBOLS_HINT` and `DATE_HINT` were added to the *shared* module. No new module-local duplicate of an existing shared concept, so #135 is not worsened.

## Verification

- `make lint` clean; full unit suite **1137 passed, 2 skipped**
- `register_batch` over all 216: **0 failures, 216 registered** — which also proves every one of the 33 has complete *and* non-extra hints, since `_validate_parameters` demands exact set equality
- No endpoint lost, duplicated, or misplaced by partitioning; every endpoint's semantics category equals its group's
- All 224 MCP `method_name`s swept through `get_expected_category`: **0 misroutes** under longest-prefix, including `get_quotes` (batch) vs `get_quote` (company) and `get_income_statement_bulk` vs `fundamental.get_income_statement`
- Both new guards mutation-tested: empty partition return, silent endpoint drop, silent semantics drop, empty catalog, and reverting the #128 fix were all caught

## Follow-up filed

**#144** — `EndpointParam.required` is dead metadata for 13 optional params. No runtime bug; the data model contradicts itself and will mislead a reader.

## Note

`Closes` does not fire on dev-base PRs here — #128 and #129 need closing by hand after merge.


MehdiZare:
@coderabbitai review

MehdiZare:
## Review findings addressed — `2656fb5`

A full review of `173ebe3..b8c8603` ran against this branch, independently re-deriving every numeric claim rather than trusting the description, sweeping **all 216 endpoints** rather than sampling, and mutation-testing the new guards.

Everything reproduced: 18 groups, 216 endpoints, 4 exclusions, 33 hints filled (batch 22 / sec 9 / transcripts 2), 13 mandatory-with-default, 13 optional-with-`required=True`, 81 endpoints carrying 167 optional params. The full-catalog sweeps came back clean on every axis that mattered — required-set == mandatory-set (**0 mismatches across all 216**), no param in both lists, no optional param at `ParamLocation.PATH` (which would have broken URL construction when omitted), 0 misrouted endpoints, 0 duplicated across groups, 0 hint drift. The 4 exclusions were independently computed as the complete collision set. And **seven source mutations were caught by the new guards**, including removing the sec exclusions, flattening the partitioning, and reverting #128 itself.

One finding was worth holding for.

### Important — nothing pinned the third-party contract the fix depends on

Giving optional params `default=param.default` rather than `None` is load-bearing: `Endpoint.validate_params` marks a param seen *before* it skips a `None`, so an explicit `None` suppresses the endpoint's own default on the way out. 64 optional params carry a non-`None` default, so getting this wrong drops `period=annual`, `limit=40` and friends off the wire on every LLM call that omits them.

But that only holds if langchain forwards fields holding explicit defaults — an implementation detail of `BaseTool._parse_input`, which has previously used the narrower `if k in tool_input` filter, in a dependency pinned `langchain-core>=1.4.9` **with no upper bound**. Every existing assertion inspects `model_fields[...].default`, never the kwargs a tool actually forwards, so if that behaviour reverted the defaults would silently stop reaching the API **with the entire suite still green**.

`test_omitted_optional_reaches_the_endpoint_with_its_declared_default` now drives a real `StructuredTool`, invokes it with only the mandatory param, and asserts on what the wrapped function received:

```
{"symbol": "AAPL"}  ->  {'symbol': 'AAPL', 'period': 'annual', 'limit': 40}
```

Verified non-vacuous — defaulting optionals to `None` fails it with `{'period': None} != {'period': 'annual'}`. The `create_parameter_fields` docstring now names the coupling so it is greppable when someone bumps the pin.

### Minor — fixed here

- **`DATE_HINT` → `AS_OF_DATE_HINT`.** The new shared constant sat one character and seven lines from the existing `DATE_HINTS` range mapping, with opposite meanings — a trap for the next person reaching for a date hint. `batch` was its only importer.
- **`SIC_CODE_HINT` dropped its `\b\d{4}\b` pattern**, which claimed any four-digit number in a query — a year, a `limit=1000`. The `sic`-anchored pattern beside it loses nothing.
- **CHANGELOG:** both breaking notes were accurate but filed as sub-bullets under `Fixed`. Anyone scanning for breaking changes reads `Changed`, so the schema-arity change and the `create_parameter_fields` signature break are promoted there.
- Design spec projected 168 → 224 guarded; as built it is 216, and the line now says why the estimate was over (the four modules add 52, not 56, and four more are held out as collisions).

### Not folded in — #144 already owns it

13 params declare `required=True` while sitting in `optional_params`, so this change lets an LLM call `get_crypto_historical`, `get_earnings_calendar` and four others with no date range, and `get_historical_social_sentiment` with no `page`.

Checked whether that creates an internal inconsistency, and it does not: `EndpointParam.required` is read **nowhere** in `fmp_data/`, and `validate_params` resolves requiredness purely by list membership — so the new schema matches runtime exactly, and this is the only self-consistent choice available here. It is still a real LLM-facing change, and which way to reconcile the flag (move those params to `mandatory_params`, or drop the flag) depends on whether FMP tolerates e.g. `historical-price-eod/full` without `from`/`to`. That is a live-API question, and #144 is the right place to settle it with a probe.

Also left alone, for the record: `_partition_by_category`'s fallback bucket parks a semantics-less endpoint in the first bucket. Zero such endpoints exist, and it is double-guarded by `test_every_endpoint_has_semantics` and `test_every_endpoint_routes_to_its_own_groups_category`. Shrinking the guarded universe instead would reintroduce the #121 failure mode.

### Verification

Run in **both** extras environments, since CI installs none by default: all-extras **1137 passed, 3 skipped**; no-extras **1020 passed, 5 skipped**. `ruff check` clean, changed files format-clean, `mypy` clean on all four changed modules, every pre-commit hook passed.

Issue #146, opened 2026-08-08
Numbers in docs/api/endpoints.md and docs/mcp/configurations.md are unguarded


MehdiZare:
Found during review of the #126/#130/#136 PR.

## What

`docs/mcp/tools.md` is fully protected: `tests/unit/test_docs_tools_sync.py` enforces its header totals, per-section counts, TOC counts, and the complete row set. That guard is why the catalog change in #130 could not leave a stale number behind.

`docs/api/endpoints.md` and `docs/mcp/configurations.md` have no equivalent. The Institutional table in `endpoints.md` currently says 24 rows and 24 is correct — verified against the module — but nothing keeps it correct. Removing `cik_mapper_by_name` took that count from 25 to 24, and only a manual edit kept the doc honest.

## Why it matters

This is precisely the drift class #123, #129 and #136 were all about: a number in a doc that nothing recomputes. The pattern the project has already converged on — assert the doc against the code — works, and is simply not applied here.

## Suggested fix

Extend `test_docs_tools_sync.py`, or add a sibling, covering:

- `docs/api/endpoints.md` — per-client endpoint tables against the actual `*_ENDPOINT_MAP` contents, not just counts. Row-set equality catches renames as well as miscounts, which is what makes the `tools.md` guard effective.
- `docs/mcp/configurations.md` — it states no counts today, so the useful assertion is different: that every file path it references exists, and every tool key it names is real and not deprecated. It currently points at `examples/mcp_configurations/` while the real path is `examples/mcp/configurations/` (fixed in the #126/#130/#136 PR), which a path-existence check would have caught.

## Scope

No runtime impact. This is about keeping documentation from quietly going stale, in a repo that has already been bitten by exactly that several times.

MehdiZare:
Fixed by #155 (merged).

`docs/api/endpoints.md` and `docs/mcp/configurations.md` now have the guard `docs/mcp/tools.md` already had. Per-client counts and per-section table rows are asserted against the code, and every `examples/...` path the configuration guide names must exist.

It found real drift immediately: the TOC claimed 47 Market Intelligence endpoints when the code and the document's own table both had 46. The row-set version also caught two intraday paths that a count-only check would have missed, when #169 corrected `historical-chart/{interval}/{symbol}` to `historical-chart/{interval}`.

Closing by hand — `Closes` did not fire on this dev-base PR.

PR #147, opened 2026-08-08
fix(mcp): one tool key per (client, method) pair (#126, #130, #136)

### LINKED ISSUE #126: MCP: key-only tool specs 'crypto_quotes' and 'forex_quotes' are ambiguous across clients

### LINKED ISSUE #130: institutional.cik_mapper_by_name is a duplicate tool that cannot accept a name
## What

`CIK_MAPPER_BY_NAME` (`fmp_data/institutional/endpoints.py`) and `CIK_MAPPER` are the same endpoint: both `path="cik-list"`, both with only `page` and `limit` as parameters. Neither takes a name.

The actual name filtering lives in the client wrapper, `InstitutionalClient.search_cik_by_name` (`fmp_data/institutional/client.py`), which notes that *"The FMP API does not support server-side name filtering"*, fetches `limit=10000` and filters locally.

LangChain tools call `client.request(endpoint, **kwargs)` directly (`fmp_data/lc/vector_store.py`), bypassing the client wrapper. So the tool generated from `cik_mapper_by_name` has no parameter through which a name can be expressed — it is a functional duplicate of `cik_mapper`.

## How it surfaced

#127 removed the `name` parameter hint from `institutional.cik_mapper_by_name` because registry validation rejects hints for parameters the endpoint does not have. That removal is correct against the endpoint as written, but it deleted the last trace of the entry's intent: a validation failure became an invisible dead tool.

## Scope

- Not in `DEFAULT_TOOLS` (`"institutional.cik_mapper_by_name" in DEFAULT_TOOLS` → `False`)
- Present in the LangChain vector-store catalog and the 224-tool MCP catalog
- The sync/async `search_cik_by_name` client methods work correctly and are unaffected

## Options

1. **Drop `cik_mapper_by_name` from `INSTITUTIONAL_ENDPOINT_MAP`.** The client method is the genuine interface; there is no distinct endpoint to model. Removes one tool from the catalog (224 → 223), so `docs/mcp/tools.md` counts and the guard tests move with it.
2. **Give the endpoint a `name` parameter** and have the client wrapper consume it for local filtering. Keeps the tool, but models a parameter the upstream API does not accept — which will read as a lie to anyone comparing against FMP's docs.
3. **Teach the LangChain tool layer to dispatch through client methods** rather than `client.request`, so wrappers with real logic are reachable. Largest change, but it is the general form of the problem: any client method that does more than pass through is invisible to LangChain today.

Option 1 is the smallest honest fix. Option 3 is worth considering separately.

Worth confirming against the live API first that `cik-list` really has no name filter — per repo convention, probe with curl rather than trusting the comment.

Found during review of #127.

### LINKED ISSUE #136: MCP catalog double-lists three company methods under singular and plural keys

MehdiZare:
Closes #126
Closes #130
Closes #136

These three issues are one question — **what does the MCP tool-key namespace guarantee?** The answer this PR implements: **one tool key per `(client, method)` pair.** Remove what is dead now; deprecate live aliases over 2.6 → 3.0.

## #130 — `institutional.cik_mapper_by_name` removed outright

Not deprecated, removed: it is not a second name for a working tool, it is a tool that cannot express the operation it claims. `CIK_MAPPER_BY_NAME` and `CIK_MAPPER` were both `path="cik-list"` with only `page` and `limit`; neither takes a name.

Confirmed against the live API on 2026-08-07 rather than assumed — `/stable/cik-list` with `name=Apple`, with `company=Apple`, and unfiltered returned **byte-identical unfiltered pages**. There is no server-side name filter, so through the LangChain tool layer (which calls `client.request` directly) the entry was a byte-for-byte duplicate of `cik_mapper`.

The real interface is untouched: sync and async `InstitutionalClient.search_cik_by_name` still fetch `limit=10000` and filter locally. Neither ever referenced the removed endpoint.

Catalog **224 → 223**; institutional 13 → 12.

## #136 — three methods were advertised under two keys each

`company.get_executives`, `get_historical_prices` and `get_intraday_prices` each appeared under two keys, and **both halves of all three pairs were in `DEFAULT_TOOLS`** — so the default server advertised three pairs of identical tools with nothing to tell them apart.

Canonical / deprecated: `historical_prices` / `historical_price`, `intraday_prices` / `intraday_price`, `key_executives` / `executives`.

Deprecated keys still resolve through 2.6 and emit a `DeprecationWarning` naming the replacement; they leave `DEFAULT_TOOLS` now (**159 → 156**) and are removed at 3.0 (**223 → 220**).

## #126 — ambiguous bare keys now fail usefully

`crypto_quotes` and `forex_quotes` are each claimed by two clients. Both tools are legitimate and neither is removed — what changes is the guarantee: **bare-key resolution is supported only for keys exactly one client claims.** An ambiguous bare key now raises an actionable error instead of a bare `RuntimeError`:

```
Tool key 'crypto_quotes' is claimed by 2 clients (alternative.crypto_quotes,
batch.crypto_quotes), so the bare key cannot be resolved. Bare tool keys work
only when exactly one client claims them; name the client explicitly using the
full '<client>.<key>' form, e.g. 'alternative.crypto_quotes'.
```

The shrinking allowlist in `test_discovered_tool_keys_are_unambiguous_globally` is replaced by a test asserting the ambiguous set is *exactly* the documented pair — a new undocumented collision still fails.

## Our own examples had the bug

Review caught that the shipped manifests `docs/mcp/configurations.md` tells users to copy listed **both** keys of a pair — `trading_manifest.py` had `historical_price` *and* `historical_prices`; `research_manifest.py` had `executives` *and* `key_executives`. Registering `trading_manifest.TOOLS` emitted 2 deprecation warnings and registered duplicate tools: the exact #136 problem, in the examples. Now:

```
trading_manifest:  23 tools, 0 deprecation warnings   (was 25 with duplicates)
minimal_manifest:   8 tools, 0 deprecation warnings
research_manifest: 37 tools, 0 deprecation warnings
```

A guard test asserts no shipped example manifest names a deprecated key.

## Also fixed from review

- `stacklevel=3` → `4`: the `DeprecationWarning` was attributed to `tool_loader.py` instead of the caller, which silently defeats any downstream `filterwarnings` rule keyed on the user's module. Verified it now attributes to the calling file.
- `get_recommended_tools()` was handing out `company.historical_price` — a key this same release deprecates.
- `fmp-mcp validate` reported a manifest naming deprecated specs as simply "valid"; it now names them and their replacements (still exit 0 — deprecated is valid, just not future-proof).
- A **3.0 breadcrumb**: `assert int(__version__.split(".")[0]) < 3` fails at the major bump with instructions, so the cleanup does not depend on memory.
- Corrected doc path `examples/mcp_configurations/` → `examples/mcp/configurations/`.

## Verification

- `make lint` clean; full unit suite **1152 passed, 2 skipped**
- Guards proven non-vacuous by injection: a new cross-client collision, a same-client duplicate key, and an ambiguity silently disappearing were each caught, then restored
- Catalog diff vs `dev` is exactly one removal — nothing else dropped, renamed or re-keyed; per-client counts moved only where expected
- `docs/mcp/tools.md` totals recomputed and enforced by `test_docs_tools_sync.py`

## Follow-up filed

**#146** — `docs/api/endpoints.md` and `docs/mcp/configurations.md` have no guard test, unlike `tools.md`. Same drift class these issues are about.

## Note

`Closes` does not fire on dev-base PRs here — #126, #130 and #136 need closing by hand after merge.


MehdiZare:
@coderabbitai review

MehdiZare:
## Review findings addressed — `b85a6e4`

A full review of `173ebe3..3fc42af` ran against this branch, independently re-deriving every numeric claim in the description rather than taking it on faith, re-probing the live API, and mutation-testing the new guards. Result: **all numeric claims reproduced exactly** (catalog 223, `DEFAULT_TOOLS` 156 unique, institutional 12, 220 at 3.0, manifests 23/8/37, catalog diff exactly one removal), the `/stable/cik-list` probe **reproduced today** — `name=Apple` and `company=Apple` still return byte-identical unfiltered pages, so removing `CIK_MAPPER_BY_NAME` outright is the honest call — and **six targeted mutations were caught by six tests**, including an ambiguity *silently disappearing*, which is the half most implementations miss.

One finding was a genuine blocker.

### Important — the deprecation cycle announced nothing to the people it exists for

`stacklevel=4` is correct for the path the test exercises and wrong for the path everyone uses. The dominant entry point is `create_app` → `register_from_manifest`, which `python -m fmp_data.mcp`, `fmp-mcp serve` and Claude Desktop all funnel through. Frame 4 is then `create_app` itself, so the warning is attributed to `fmp_data.mcp.server` — a library module — and Python's default filter chain (`default::DeprecationWarning:__main__`, then `ignore::DeprecationWarning`) discards anything not attributed to `__main__`.

Reproduced before fixing. A real `create_app(["company.historical_price"])` under default filters:

```
--- create_app under DEFAULT filters ---
--- done ---            # nothing emitted
```

Under `-W default::DeprecationWarning` it appears, attributed to `server.py:75`. `test_warning_is_attributed_to_the_caller_not_the_library` passed only because it calls `register_from_manifest` directly, so it gave false confidence about the one case that matters.

**Fix:** the `warnings.warn` stays — it is the right contract for direct callers and the only thing a downstream `filterwarnings` rule can match — and the same message now also goes out at `WARNING` on the `fmp_data.mcp.tool_loader` logger, which lands on the server's stderr where MCP hosts collect it. No stacklevel could have solved this anyway: the offending key usually lives in a manifest *data* file, not in a stack frame.

`test_deprecation_survives_pythons_default_warning_filters` reproduces those exact filters, asserts the warning *is* swallowed (so it keeps proving something) and that the log record still carries key, replacement and the 3.0 target. Removing the `logger.warning` fails it. `test_canonical_keys_are_silent_on_both_channels` pins that the new channel does not turn every canonical key into noise.

### Important — the `DEFAULT_TOOLS` reduction is breaking, and the CHANGELOG said otherwise

It claimed *"only manifests naming a deprecated key explicitly are affected"*. Not true of someone running the default server whose saved prompt names the tool `executives` — that tool disappears on upgrade, and they cannot be warned in code, because they never name the key so no resolution happens and there is no call path on which a warning could fire.

Keeping the change: advertising two identical tools per method is the defect #136 reports, and it cannot be fixed while both remain advertised. #136 itself lists this as option 2. What changed is that the CHANGELOG now says so plainly — flagged **breaking for default-server users in a minor release**, with the rename to apply and the manifest escape hatch for anyone who needs the old names through 2.6.

### Minor — fixed here

- `docs/api/endpoints.md` TOC said Institutional (25 endpoints); the section heading and the real count are 24. Exactly the doc-drift class this PR is otherwise careful about.
- `docs/mcp/configurations.md` stopped short of the failure a reader hits next: naming **both** halves of a colliding pair still fails under the default `key` name style. Now points at the `FMP_MCP_TOOL_NAME_STYLE=spec` tip already on the page.
- `test_catalog_registration.py` docstring still justified itself with `intraday_price` sitting in `DEFAULT_TOOLS`, which this branch changed.
- Design spec said catalog → 218 in 3.0; it is 220.
- Dropped a redundant local `import warnings`.

### Filed rather than folded in

- **#148** — `fmp-mcp generate` was left out of the deprecation awareness that `list` and `validate` gained, and emits a manifest that also cannot start a server (223 tools including 3 deprecated; registration then fails on `crypto_quotes`/`forex_quotes`). The deprecated-spec half is a one-liner, but the duplicate-key half is a separate design question about the tool-name contract.
- **#149** — bare-key resolution is now implemented three times (`_resolve_tool_spec`, `fmp-mcp validate`, the example-manifest guard). They agree today with nothing keeping them in agreement; drift would let `validate` bless manifests the loader refuses. Fixing it properly means splitting resolution from announcement, which reshapes the `tool_loader` API.

Left alone deliberately: `_warn_if_deprecated` fires before the full spec is known to exist. Harmless today — all three deprecated specs are real — and adding a catalog check there is scope this PR shouldn't carry.

### Verification

`1153 passed, 3 skipped`; `ruff check`, `ruff format --check` and `mypy fmp_data/mcp` all clean; every pre-commit hook passed. CI green on `b85a6e4` (17/17).

PR #150, opened 2026-08-08
fix(lc): consolidate hints, raise on vector-store failure, hide deprecated endpoints (#135, #133, #137)

### LINKED ISSUE #133: create_vector_store swallows the fail-loud errors setup_registry now propagates
## Context

#127 changed `setup_registry()` so that non-validation errors propagate instead of being swallowed by a blanket `except Exception` — the "loud failure" half of the #121 fix. The intent: a client group that cannot load at all should not look like a group that legitimately has nothing in it.

## Problem

The only production caller cancels it out. `create_vector_store()` wraps the whole body — including the `setup_registry(client)` call — in a blanket handler:

```python
# fmp_data/lc/__init__.py:255-257
except Exception as e:
    logger.error(f"Error creating vector store: {e!s}")
    return None
```

So an `ImportError`/`AttributeError` raised while `register()` lazily builds validation rules via `_ensure_validation_initialized()` (`fmp_data/lc/registry.py:405-422`) still reaches a library user as a bare `None`. Nothing that #127 made propagate is observable to any in-repo consumer.

The net behavioral change is arguably the opposite of what was intended: a group that fails to load used to degrade the store, and now kills it entirely — reported only by a log line plus a `None` the caller has to remember to check.

## Why this wasn't fixed in #127

Narrowing that catch changes the public behavior of a documented, exported function: callers that today branch on `if store is None` would start seeing exceptions. That needs a deliberate decision about the error contract, not a drive-by change inside a fix PR.

## Options

1. Let a typed subset out — e.g. `ImportError`/`ConfigError` propagate, everything else keeps returning `None`.
2. Introduce a dedicated `VectorStoreCreationError` and raise it, deprecating the `None` return over a release.
3. Keep `None` but return a result object carrying the failure reason.

## Notes

- MCP is unaffected: `setup_registry` is never called on any MCP path — `fmp_data/mcp/tool_loader.py` and `discovery.py` read the `*_ENDPOINTS_SEMANTICS` tables directly.
- Related: `setup_registry()` inspects `register_batch()`'s returned failure dict but does not propagate it — its return type is still just `EndpointRegistry`, so a programmatic consumer can only learn about skipped endpoints by parsing logs. Worth settling in the same design pass.

Found during review of #127.

### LINKED ISSUE #135: Consolidate duplicated ParameterHint constants into fmp_data.lc.hints

### LINKED ISSUE #137: LC/MCP: deprecated endpoints are still selectable in the vector store
## What

Three intelligence endpoints are deprecated and return `[]` without calling upstream, but they remain in `INTELLIGENCE_ENDPOINTS_SEMANTICS`, so `EndpointVectorStore` still indexes them and a semantic query can select one:

- `stock_news_sentiments`
- `earnings_confirmed`
- `earnings_surprises`

An LLM that picks one gets an empty success — indistinguishable from "no data for your query".

## Already mitigated

- All three descriptions lead with `DEPRECATED` and name their replacement
- `stock_news_sentiments` had its `related_terms` / `example_queries` retargeted so it stops competing for live sentiment queries
- All three are out of `DEFAULT_TOOLS`, so a default MCP server does not register them

What remains is that the LangChain vector store has no notion of deprecation — it indexes whatever is registered.

## Proposed fix

Add a `deprecated: bool = False` field to `EndpointSemantics` and filter on it where endpoints are indexed (`EndpointVectorStore.add_endpoints` / `get_tools`), rather than deleting the entries.

Why this shape:

- It handles every future deprecation uniformly instead of case-by-case description edits
- It keeps the tool keys resolvable for anyone loading them through an explicit manifest, so nothing breaks without a deprecation cycle
- The catalog count stays 224, so `docs/mcp/tools.md` and its guard test do not move

Deleting the three keys outright is the alternative, but that removes public tool keys with no deprecation window and drops the catalog to 221 — the same class of change as #130 and #136, which are deferred for that reason.

## Scope

- `fmp_data/lc/models.py` — new field
- `fmp_data/lc/vector_store.py` — filter at index and/or selection time
- `fmp_data/intelligence/mapping.py` — mark the three
- Tests: a deprecated endpoint must not be returned by `get_tools`

## Context

Raised by CodeRabbit on release PR #113 (`fmp_data/intelligence/mapping.py:578`), deferred out of the release. See https://github.com/MehdiZare/fmp-data/pull/113#discussion_r3737008724

MehdiZare:
Closes #135
Closes #133
Closes #137

## #135 — ten hint concepts had divergent definitions

Ten `ParameterHint` names were bound in more than one module at once with *different* contents — `SYMBOL_HINT` in six modules, `CIK_HINT` and `PAGE_HINT` in four each. These drive natural-language parameter extraction, so the same user phrasing resolved on one tool and not another, and nothing compared them.

Each concept now has exactly one definition in `fmp_data.lc.hints`, imported by the domain modules. Per the agreed policy the union was taken, **except where a union would produce wrong extraction rather than merely a longer list**. Every such case is recorded in `hints.py`:

- **Unanchored patterns dropped, not unioned** — `[A-Z]{1,5}` matched inside longer tokens (`NASDAQ` → `NASDA`), bare `(\d{10})` truncated CIKs, `p(\d+)` turned `sp500` into page 500. The last was narrowed to `\bp(\d+)\b` rather than dropped, so the shorthand still works.
- **US-order `MM/DD/YYYY` not unioned** — every date param in the catalog takes ISO, so advertising it would suggest an unsendable format.
- **`QUARTER_HINT` dropped `Q1`/`Q4` examples** — every `quarter` param in the catalog is `ParamType.INTEGER`; `Q1` would invite a 400.
- **`PERIOD_HINT` split from a new `PERIOD_LENGTH_HINT`** — `period` is a reporting frequency (`annual`, `quarter`, all `ParamType.STRING`); `periodLength` is a numeric lookback (all `ParamType.INTEGER`). Merging would offer `annual` as an RSI window and `50` as a statement period. A test pins them apart.

A guard test asserts no two modules define a `ParameterHint` under the same name with different contents, with floors on modules scanned, bindings found, and shared names.

## #133 — `create_vector_store` swallowed the errors #121 made loud

The function wrapped everything, including `setup_registry`, in `except Exception: return None`, cancelling out the fail-loud behaviour added in #121.

`None` carried no information: there was exactly one caller-reachable `return None` — the blanket except. So `None` unconditionally meant "something threw", and the caller could not tell what.

**Breaking change.** `create_vector_store` now raises `VectorStoreCreationError` (carrying `cause` and the per-endpoint `failures` dict) and its return type narrows to `EndpointVectorStore`. There is no `None` path and no compatibility shim — since `None` never carried information, there was nothing for one to preserve.

```python
# before
store = create_vector_store(...)
if store is None:
    ...   # why? no way to tell

# after
try:
    store = create_vector_store(...)
except VectorStoreCreationError as exc:
    exc.cause      # the underlying exception
    exc.failures   # {endpoint name: validation error}
```

`setup_registry`'s return widens so `register_batch`'s failures are programmatically reachable rather than log-only — #133's closing note. README, `LLM.md` and the integration tests are migrated.

## #137 — deprecated endpoints were still selectable

`stock_news_sentiments`, `earnings_confirmed` and `earnings_surprises` return `[]` without calling upstream, but stayed in the vector store — so an LLM could pick one and get an empty success indistinguishable from "no data for your query".

New `EndpointSemantics.deprecated` flag, filtered at index time (`add_endpoint`/`add_endpoints`), at retrieval (`get_tools`), and in `search()`. Entries are **not** deleted: tool keys stay resolvable for explicit manifests and the MCP catalog stays at 223.

Selection-time filtering matters because a store persisted by an older release has these baked into its index; filtering only at index time would leave those users where #137 found them.

This is a **different mechanism** from `DEPRECATED_TOOLS` (#147), which is about duplicate *names* for a live method. Documented in four places plus a test asserting the two sets are disjoint, so they do not get merged by mistake later.

## Review findings fixed

- **`search()`'s filter was untested** — the primary LLM-facing path. Deleting it left all 66 `tests/unit/lc` tests green. Now guarded against the real registry (a stub cannot reach `search`, since `SearchResult.info` is typed `EndpointInfo`), and the guard was verified to fail when the filter is removed.
- **Four undocumented content drops unioned back** — `SYMBOL_HINT`'s `company symbol`/`stock symbol` (18 endpoints), `CIK_HINT`'s three extra examples and two natural names (12), `AS_OF_DATE_HINT`'s two examples (9). These feed `get_embedding_text` and `generate_description`, so dropping them shifted embedding text and tool descriptions with no recorded reason.
- **Accidental-deprecation guard widened** from the intelligence table to every semantics table, with a floor — another domain could previously acquire `deprecated=True` and vanish from the vector store unnoticed. Verified by flipping a `market` endpoint.
- **"nine" → "ten"** consolidated concepts in the CHANGELOG and module docstring; both already enumerated ten.

## Verification

- `make lint` clean; full unit suite **1190 passed, 2 skipped**
- Hint duplication guard mutation-tested: injecting a divergent `PAGE_HINT` fails and names exactly which modules disagree
- Deprecation filters mutation-tested at all three sites; removing any one fails a test
- No extraction pattern silently lost — every dropped pattern is either subsumed by a broader replacement or recorded as a deliberate conflict
- CHANGELOG migration snippet verified by running it against a forced `ImportError`

## Note

The implementer hit a session limit after committing and never filed its report, so the three original commits were verified independently rather than self-reported.


MehdiZare:
## Code review round: three independent reviewers, findings addressed in 1fb7fa0

Reviewed `d43b7de..1a46e6a` from three angles — hint-consolidation correctness (#135), error contract + deprecation filtering (#133/#137), and test/docs quality. **No Critical findings.** Every mutation claim in the PR body was independently reproduced: all six mutations were caught, none survived. The "None never carried information" justification for the breaking change was re-derived from the base tree rather than taken on faith, and the CHANGELOG migration snippet was executed.

### Fixed in this PR

**1. `PERIOD_HINT`'s union advertised values the endpoints reject.** The catalog enforces three different `valid_values` sets under the one `period` parameter. `examples` is rendered verbatim into the tool description, while `EndpointParam.validate_value` rejects anything outside `valid_values` — and `create_parameter_fields` never surfaces the enum to the model. So the union told three `company` endpoints to send `FY`/`Q1` (a regression this PR introduced) and six bulk endpoints to send `annual`/`quarter` (pre-existing). This is exactly the failure the PR correctly refused for `QUARTER_HINT`, applied inconsistently one hint over.

Split into `PERIOD_HINT` (annual/quarter), `PERIOD_WITH_FISCAL_HINT` (+ FY/Q1–Q4) and `FISCAL_PERIOD_HINT` (fiscal only), one per value set. A catalog-wide scan now reports **0 violations across 215 endpoints**, down from 9.

**2. `search` shrank its own answer on a stale index.** Deprecated entries persisted by an earlier release occupy top-`k` slots, so filtering after the fetch returned `k-1` live endpoints — under-recall in precisely the case selection-time filtering exists for. `search` now widens the fetch while filtered entries are displacing live ones, then truncates to `k`. Threshold drops deliberately do not trigger a refetch: that is a genuine relevance cut.

**3. Two tests ran in no CI job at all.** `test_mcp_catalog_still_advertises_the_deprecated_keys` and `test_lc_deprecation_and_mcp_deprecated_tools_are_separate` need `mcp`, and `tests/unit/lc/` is skipped without `langchain` — which no `mcp` job installs. The `mcp-server` job runs only `tests/unit/test_mcp.py`. Meanwhile the CHANGELOG asserted "a test asserts the two sets stay disjoint": true, but never executed. Moved to `tests/unit/test_mcp.py`; **verified passing in an mcp-only venv with no langchain installed**. The same defect class commit `1a46e6a` fixed for #133, reintroduced one file over.

**4. The accidental-deprecation guard was invisible to the core matrix.** The four data-only #137 tests need no extra, so they now live in a top-level `tests/unit/test_deprecated_endpoint_flags.py` and run in the default no-extras jobs, not only in `langchain`.

**5. New guard: `test_hint_examples_are_within_valid_values`.** Walks all 215 registered endpoints and fails when a hint advertises a value its parameter rejects — the class of defect in (1), which no test could previously see. Floors on endpoints scanned and constrained params found; mutation-tested by re-merging the period hints.

Also: `PAGE_HINT`'s main pattern anchored (`webpage3` read as page 3 — the widest of its four copies); `add_endpoints` returns the indexed count so `create_new_store` stops logging 215 offered as if 212 were stored; `VectorStoreCreationError.__str__` appends the skipped endpoints, following `RateLimitError`; error-contract floor raised `>100` → `>=200` so a re-regression to #129's 168 cannot pass; one vacuous-if-empty assertion given a non-empty guard; registry count corrected 216 → 215 in the CHANGELOG (wrong at base too); the `fmp_data.company.hints` import-path removal now documented; and two "subsumed by" comments corrected to "deliberately narrowed" — the anchored patterns match strictly *less*, which is the point, and the comments exist so the decision can be re-derived.

Each of the three new behaviours was mutation-tested: single-round fetch, re-merged period hints, and `len(names)` for the indexed count all fail their guard.

### Deferred to follow-up issues

- **#156** — derive tool examples from `valid_values` and surface `valid_values` as a `Literal` in the generated schema. That is the root fix for (1); the three-way split manages the symptom. With it, the period hints could collapse back to one.
- **#157** — the duplication guard is keyed on binding *name*, so `technical`'s `FROM_DATE_HINT`/`TO_DATE_HINT` (same concept as `DATE_HINTS`, different contents) stay invisible. Folding them in changes nine indicator tools and deserves its own before/after.
- **#158** — `fmp-mcp list` marks deprecation only from `DEPRECATED_TOOLS`, so the three dead endpoints print unmarked and users copy them into manifests.
- **#159** — `_is_deprecated(info: Any)` plus `getattr(..., False)`: rename the field and filtering silently stops.

### Verification

- All extras: **1196 passed, 3 skipped**
- Core-only venv (httpx/pydantic/tenacity, no langchain/faiss/mcp): **1037 passed, 11 skipped**
- mcp-only venv (simulating the `mcp-server` job): the two moved tests **pass**
- `ruff check`, `ruff format --check`, `mypy` clean

Two unrelated markdown files that `ruff format` reflowed (a dated plan doc and an MCP configuration example) were reverted rather than carried in.

PR #151, opened 2026-08-08
fix(mcp): one pure resolver, and a startable generated manifest (#149, #148)

### LINKED ISSUE #148: MCP: `fmp-mcp generate` emits a manifest that cannot start a server
Found during review of #147.

`generate_manifest(tools=None)` (`fmp_data/mcp/cli.py:170`) writes every catalog spec verbatim. The generated manifest is unusable:

```
generated tools: 223
deprecated included: ['company.executives', 'company.historical_price', 'company.intraday_price']
registration FAILS: Duplicate tool keys detected for MCP tool names: crypto_quotes, forex_quotes.
```

Two distinct problems, deliberately left out of #147 to keep that PR narrow:

## 1. `generate` was not made deprecation-aware, unlike `list` and `validate`

#147 taught `fmp-mcp list` to mark deprecated rows and `fmp-mcp validate` to report deprecated specs alongside their replacements. `generate` was missed, so it hands the user a manifest seeded with the three keys that same release deprecates — and then warns them about it on every server start.

Fix is a one-liner: filter `DEPRECATED_TOOLS` out of the default spec list.

## 2. Full specs still collide under the default `key` tool-name style

The deeper half. Under `FMP_MCP_TOOL_NAME_STYLE=key` (the default), the advertised tool name is the bare key, so `alternative.crypto_quotes` and `batch.crypto_quotes` both want to be called `crypto_quotes` and `_validate_tool_names` refuses the pair. This is the other face of #126: #147 fixed *bare-key resolution* for ambiguous keys, but naming both full specs in one manifest is still a hard failure.

That is arguably correct behaviour — two tools cannot share one advertised name — but it means "every tool in the catalog" is not a expressible manifest under the default style. Options:

1. `generate` emits `FMP_MCP_TOOL_NAME_STYLE=spec` guidance (or a comment) when the tool set contains a collision.
2. `generate` defaults to excluding one side of each collision, documented in the file header.
3. Auto-qualify only the colliding names under `key` style, leaving the rest bare.

Option 3 changes the tool-name contract and deserves its own discussion; 1 and 2 are cheap.

## Scope

Pre-existing for the duplicate-key half; newly inconsistent for the deprecated-spec half. Neither affects the default server (`DEFAULT_TOOLS` has no collisions and, as of #147, no deprecated keys) — only users of `fmp-mcp generate` with no `--tools` filter.

### LINKED ISSUE #149: MCP: bare-key resolution logic is written three times and can drift
Found during review of #147.

The rule "a bare tool key resolves only when exactly one client claims it" is now implemented independently in three places:

| Location | Role |
|---|---|
| `fmp_data/mcp/tool_loader.py:88` `_resolve_tool_spec` | the real resolver, used at registration |
| `fmp_data/mcp/cli.py:219-230` `_classify_manifest_entries` | `fmp-mcp validate`'s reimplementation |
| `tests/unit/test_mcp.py:690` | the example-manifest guard's reimplementation |

They agree today. They are not kept in agreement by anything.

## Why it matters

Validator drift is the bad case: if the resolution rule changes and only `tool_loader.py` is updated, `fmp-mcp validate` blesses manifests that the loader then refuses at server start. A validator that disagrees with the thing it validates is worse than no validator, because it converts a startup error into a trusted green check.

## Why it wasn't fixed in #147

`validate` cannot simply call `_resolve_tool_spec` today: resolution emits a `DeprecationWarning` (and, as of #147, a `logger.warning`) as a side effect, so a validator looping over entries would spray spurious deprecation noise for keys the user is merely *checking*, not loading.

## Suggested direction

Split resolution from announcement — e.g. a pure `resolve_tool_spec(spec, key_to_spec) -> Resolution` that returns whether the spec is deprecated and what it resolves to, with `_warn_if_deprecated` called only by the registration path. Then `validate` and the test guard both consume the pure function and drift becomes impossible.

This is the strongest architectural smell in the #147 diff, but fixing it properly means reshaping the tool_loader API, which is more than that PR should carry.

MehdiZare:
Closes #149
Closes #148

Both were filed during review of #147 and deliberately left out of it.

## #149 — the resolution rule was written three times

"A bare tool key resolves only when exactly one client claims it" was implemented independently in the loader (`_resolve_tool_spec`), in `fmp-mcp validate` (`_classify_manifest_entries`), and again in a test guard. They agreed, but nothing kept them agreeing.

**Validator drift is the failure that matters**: change the rule, update only the loader, and `fmp-mcp validate` starts blessing manifests the loader refuses at server start — turning a startup error into a trusted green check.

The reason `validate` could not just call the resolver was that resolution emitted a `DeprecationWarning` *and* a `logger.warning` as side effects, so a validator looping over entries would spray deprecation noise for keys the user was merely checking.

Resolution is now split from announcement. A pure `resolve_tool_spec(spec, key_to_spec) -> Resolution` reports every outcome with **no side effects**:

```
company.historical_price     resolved=True  deprecated=True  replacement=company.historical_prices
crypto_quotes                resolved=False deprecated=False candidates=2
company.profile              resolved=True  deprecated=False
nonsense.key                 resolved=False deprecated=False
side-effect warnings: 0
```

`_warn_if_deprecated` is now called only by the registration path. `validate` and the test guard both consume the pure function, so the three implementations become one and drift is structurally impossible.

## #148 — `fmp-mcp generate` emitted a manifest that could not start a server

Before:
```
generated tools: 223
deprecated included: company.executives, company.historical_price, company.intraday_price
registration FAILS: Duplicate tool keys detected for MCP tool names: crypto_quotes, forex_quotes
```

Two causes. `generate` was never made deprecation-aware though #147 taught `list` and `validate` to be, so it seeded users with the three keys that same release deprecates. And under the default `key` tool-name style, `alternative.crypto_quotes` and `batch.crypto_quotes` both want to be advertised as `crypto_quotes`, which `_validate_tool_names` refuses — so "every tool in the catalog" was not an expressible manifest.

The issue's cheapest option was guidance in a header comment. That documents the problem without fixing it, so this goes further: `generate` excludes the deprecated keys and one side of each collision, and **reports exactly what it dropped and why**:

```
Total tools: 218
Excluded to keep the manifest startable: 2
  batch.crypto_quotes (kept alternative.crypto_quotes, both advertised as 'crypto_quotes')
  batch.forex_quotes (kept alternative.forex_quotes, both advertised as 'forex_quotes')
```

223 → 218: three deprecated keys and two collision losers. The excluded tools remain loadable by naming them explicitly, or by setting `FMP_MCP_TOOL_NAME_STYLE=spec`.

Auto-qualifying colliding names under `key` style (the issue's option 3) is deliberately **not** done here — it changes the tool-name contract and deserves its own discussion.

## Acceptance test

The bar for #148 was behavioural, not cosmetic: a manifest from `generate` with no `--tools` filter must actually register. Verified for both modes:

```
no --tools filter: 218 tools | REGISTERS CLEANLY | deprecation warnings: 0
defaults:          218 tools | REGISTERS CLEANLY | deprecation warnings: 0
```

## Verification

- `make lint` clean; full unit suite **1178 passed, 2 skipped**
- Pure resolver confirmed side-effect-free across resolved / deprecated / ambiguous / unknown
- A guard asserts `validate` and the loader agree, so a fourth reimplementation fails the build


MehdiZare:
## Review round: findings addressed

Two reviewers went over this branch (CodeRabbit auto-skips PRs based on `dev`). The branch was also rebased onto `dev` after #150 merged — `CHANGELOG.md` and `tests/unit/test_mcp.py` conflicted and were resolved by hand; a reviewer verified independently that the CHANGELOG diff against `dev` is purely additive (+12/−0) and that every removed line in `test_mcp.py` is matched by a replacement inside the same test body, so nothing was dropped from either side.

### Fixed here (ee4cc79)

**`generate` dropped deprecated specs silently, while three places claimed otherwise.** `_startable_catalog` removed them by set subtraction and appended no note, so the header explained the two collision losers and said nothing about `company.executives`, `company.historical_price`, `company.intraday_price`. A user diffing `fmp-mcp list` (223) against their manifest (218) saw two accounted-for absences and three unexplained ones — in the feature whose selling point is that nothing disappears silently. The `generate_manifest` docstring, `docs/mcp/configurations.md` and the CHANGELOG all asserted every exclusion was named.

Deprecations now get their own header section and stdout line:

```
Omitted as deprecated: 3
  company.executives (deprecated; use company.key_executives, included below)
  company.historical_price (deprecated; use company.historical_prices, included below)
  company.intraday_price (deprecated; use company.intraday_prices, included below)
Excluded to keep the manifest startable: 2
  batch.crypto_quotes (kept alternative.crypto_quotes, both advertised as 'crypto_quotes')
  batch.forex_quotes (kept alternative.forex_quotes, both advertised as 'forex_quotes')
```

Listed separately from the collision drops because the remedy differs — a collision loser needs `FMP_MCP_TOOL_NAME_STYLE=spec`, a deprecated key needs a rename. `included below` is computed against the post-collision list rather than asserted, so the note cannot lie if a replacement is ever itself dropped.

The new guard is exhaustive rather than by example: every spec in `catalog - manifest` must appear in the header, so a future exclusion rule cannot ship without its explanation.

**`_build_key_to_spec` → `build_key_to_spec`.** It builds the only argument the new public `resolve_tool_spec` takes, and this PR is what first gives it a caller outside `tool_loader` — `cli.py` did not import the module at all on `dev`. Leaving it private would have shipped a public function reachable only by reaching into a private name. Renaming a private helper breaks no documented contract.

**A stub guard for the loader**, mirroring the one `validate` already had. These catch different regressions and neither subsumes the other: the equivalence guard catches a loader that *diverges*, while an *identical* reimplementation agrees on every input and only stubbing notices. Confirmed by mutation — the def-time-capture mutation failed the new stub test while `test_registration_path_matches_the_pure_resolver_everywhere` correctly passed.

**A guard on the default-yield path.** `--tools batch.crypto_quotes` (the side `_startable_catalog` normally drops) keeps the explicit pick, and the default it displaced is named in the header. Previously verified-by-hand but unpinned.

### Mutation testing

12 mutations, 12 caught. The four #149 guards: purity broken → caught; `validate` diverging → caught; loader diverging → caught; `validate` reimplemented identically → caught by the stub test only. The #148 guards: deprecated filter removed, collision drop removed, full-spec catalog check removed, explicit selection thinned → each caught. The four guards added in this round: notes not appended → caught by both new header tests; note omitting the replacement → caught; loader bypassing the public seam → caught; default overruling the caller → caught.

### Filed rather than fixed

**#160** — `fmp-mcp generate --tools profile` calls a valid bare key unknown and writes an empty manifest with exit 0. Pre-existing on `dev`, but newly inconsistent now that a single shared resolver handles both entry forms everywhere else; `_validated_specs` is the fourth call site that still does its own membership test. Left out to keep this PR reviewable, and it wants a decision on whether an explicit selection resolving to nothing should exit non-zero.

### Considered, not acted on

`resolve_tool_spec` rebuilds a catalog set per dotted call (~49k set inserts across a 218-entry manifest). Immaterial at this catalog size and not worth complicating the pure function for; noted only against order-of-magnitude growth.

### Verification

`1214 passed, 3 skipped` (up 4 with the new guards); `ruff check` + `ruff format --check` clean; `mypy` clean over 121 source files. The behavioural acceptance test still holds: a no-filter generated manifest registers all 218 tools under `FMP_MCP_TOOL_NAME_STYLE=key` with `DeprecationWarning` promoted to error.

MehdiZare:
## Review round 2: CLI findings addressed

A second reviewer went over the CLI surface by running it rather than reading it. Findings in `8ada991`.

### Fixed here

**`fmp-mcp validate` could not see name clashes — the one registration failure that is not a property of any single entry.** #149 made the per-entry rule shared; #148 taught `generate` to refuse a manifest whose tools advertise the same name twice. `validate` learned neither, so the very file `generate` warned about validated without a word and then died at server start with `Duplicate tool keys detected`. That asymmetry is new in this PR, which makes it this PR's to close.

```
$ fmp-mcp validate clash.py          # ["alternative.crypto_quotes", "batch.crypto_quotes"]
Warning: 'crypto_quotes' is claimed by two different tools (alternative.crypto_quotes,
batch.crypto_quotes), which fails to register under the default FMP_MCP_TOOL_NAME_STYLE=key.
Set FMP_MCP_TOOL_NAME_STYLE=spec to serve both, or drop one side.

$ fmp-mcp validate dupe.py           # ["profile", "company.profile"]
Warning: 'profile' is listed more than once (profile, company.profile); registration
refuses a repeated tool name under either name style. Drop all but one.
```

Two details worth stating. Entries are **resolved before being compared** — `profile` and `company.profile` are the same tool and registration refuses the pair, but they are different strings and comparing them as written sees nothing. And the two causes are reported **separately**, because the advice differs: two genuinely different tools claiming one name are servable with `FMP_MCP_TOOL_NAME_STYLE=spec`, while one tool named twice is refused under *either* style and no name style helps. The mutation that collapses the two cases produces exactly the misleading advice, and fails.

**A deprecated key named explicitly in `--tools` is kept but reported.** An explicit ask is honoured, as with collisions — but silently, `--tools` would have been the only CLI path that puts a deprecated key in a manifest without saying so, conspicuous beside a default path that now explains every deprecation it drops.

**`_add_defaults` skips deprecated specs.** Dead defence today — the tables are disjoint, verified — but it runs *after* `_startable_catalog` has stripped deprecations, so a future deprecated entry in `DEFAULT_TOOLS` would be silently reinstated into the manifest the same release just cleaned. Guarded by injecting the overlap.

Plus: generated manifests end with a newline; an unknown `--tools` entry points at `fmp-mcp list`; and CHANGELOG:55 no longer names the pre-rename `_build_key_to_spec` two lines from where line 104 announces the rename.

### Filed rather than fixed

**#161** — `validate` still exits 0 for manifests registration refuses (unknown, ambiguous, and now the clashes above). Pre-existing on `dev`, and making it exit non-zero changes the CLI's contract for anyone with `fmp-mcp validate` in a pipeline. That is the right end state and it should be announced as a breaking change, not slipped in as a side effect of a resolver refactor. The issue proposes keeping deprecated as a report rather than a failure, and flags the `1 tools` pluralisation bug in the same line.

**#160** — `generate --tools profile` calls a valid bare key unknown and writes an empty manifest.

### Independently verified

Both reviewers checked the CHANGELOG's numbers against live code and I re-ran it myself: catalog **223**, `DEFAULT_TOOLS` **156**, `DEPRECATED_TOOLS` **3**, collisions exactly **2**, 223−3−2=**218**, LC groups **18**, LC registered endpoints **215**. That last one confirms the rebase resolution was right — this branch carried a stale **216** from before #150 merged, and 215 is what the code actually produces.

A reviewer also reproduced the "before" state to confirm the CHANGELOG quotes it verbatim: registering all 223 specs still raises `Duplicate tool keys detected for MCP tool names: crypto_quotes, forex_quotes`. And the acceptance test was re-run under **both** name styles — `key` and `spec` — 218 listed, 218 registered, 218 unique names, zero deprecation warnings, against a real `FMPDataClient`.

### Mutation testing

17 mutations across both rounds, 17 caught. This round: `validate` blind to clashes → caught; comparing entries as written instead of resolved → caught; collapsing duplicates into collisions → caught (with the wrong advice, which is the point); defaults reinstating a deprecated spec → caught; `--tools` silent on a deprecated key → caught.

### Verification

`1221 passed, 3 skipped`; `ruff check` + `ruff format --check` clean; `mypy` clean over 121 source files. All 17 CI checks were green on `ee4cc79` and are re-running on `8ada991`.

MehdiZare:
## Review round 3: a false positive in round 2's own code

The CLI verifier re-ran against `8ada991` and found two defects in the clash reporting I added last round. Both fixed in `96e5d54`.

### Fixed here

**The clash check ignored `FMP_MCP_TOOL_NAME_STYLE`.** It computed advertised names under the default `key` style unconditionally, so:

```
$ FMP_MCP_TOOL_NAME_STYLE=spec fmp-mcp validate clash.py
Warning: 'crypto_quotes' is claimed by two different tools (...). Set FMP_MCP_TOOL_NAME_STYLE=spec ...
```

It told a user to set the variable they had already set. That is the worst shape a false positive can take: it fires *only* on the person who followed the earlier advice, and what it advises is the thing they did.

Names now come from the effective style. Under `spec` the advertised name is the whole spec, so the two `crypto_quotes` tools are genuinely distinct and draw nothing. The duplicate case survives the change, as it must — `["profile", "company.profile"]` is one tool listed twice however names are derived — and that pair is exactly what the suppression must not swallow, so it is pinned by its own test rather than left to follow from the other.

```
$ FMP_MCP_TOOL_NAME_STYLE=spec fmp-mcp validate clash.py     # silent
$ FMP_MCP_TOOL_NAME_STYLE=spec fmp-mcp validate dupe.py
Warning: 'company.profile' is listed more than once (profile, company.profile) -- these are
the same tool. No tool-name style separates them; drop all but one.
```

**The duplicate message asserted something the loader does not do.** It said registration "refuses a repeated tool name under either name style". `_validate_tool_names` returns early for any non-`key` style, so under `spec` the loader accepts the manifest and the second registration silently shadows the first — verified. The message no longer promises a refusal it cannot deliver. The remedy was right and is unchanged; only the claim about what enforces it is gone.

**`validate`'s unknown-tool message now points at `fmp-mcp list`**, matching what `generate` gained last round. It was the same dead end in the other command.

### Filed rather than fixed

- **#162** — the loader skips duplicate detection entirely under `spec` style. Note the direction: #149 was about `validate` blessing what the loader refuses; this is the loader accepting what `validate` warns about. Same class of divergence, opposite sign. The early return conflates *collision* (two specs, one name — genuinely impossible under `spec`) with *duplicate* (one spec twice — unaffected by style).
- **#163** — `fmp-mcp list` is not copy-pasteable in two of its four formats. `table` truncates the spec column, so `company.historical_price` and `company.historical_prices` render identically and the `[DEPRECATED -> ...]` marker is cut off. `tree` prints method names, so the deprecated key `company.historical_price` never appears at all — the row is labelled with the *replacement's* method name. Only `list` and `json` are complete.
- **#164** — there are two deprecation mechanisms and `generate` honours one. Three endpoints carry `EndpointSemantics.deprecated` (they return `[]` without calling upstream, one describes itself as "DEPRECATED and non-functional") and are still included. #137 deliberately kept them resolvable, so omitting them from `generate` is a decision, not a patch.

### Full issue list from this review

| Issue | What |
|---|---|
| #160 | `generate --tools profile` calls a valid bare key unknown, writes an empty manifest, exits 0 |
| #161 | `validate` exits 0 for manifests registration refuses |
| #162 | Loader skips duplicate detection under `spec` style |
| #163 | `list` table/tree output is not copy-pasteable |
| #164 | `generate` honours one of two deprecation mechanisms |

### Verification

`1223 passed, 3 skipped`; `ruff` and `mypy` clean. 18 mutations across three rounds, 18 caught — this round's being the one that ignores the name style, which resurrects the false positive.

The verifier also confirmed, at `8ada991`: zero false positives on clean manifests and on the fresh 218-tool generated manifest (stderr empty in both), the newline fix, the deprecated-`--tools` warning, stream hygiene (warnings to stderr, verdict to stdout) across ~20 invocations with no stack traces, `--format json` parsing with 223 entries and a per-entry `deprecated` field, and the resolver's five statuses over the live catalog including the new catalog check on qualified specs.

MehdiZare:
## Parity check: `validate` vs `register_from_manifest`

Final verification at `96e5d54`. Both sides were run under the same environment for every combination of tool-name style (`key` / unset / `spec` / an invalid value / the messy `" SPEC "`) against a colliding manifest, a duplicate manifest, and a clean one — `register_from_manifest` driven with a stub server.

| Style | Manifest | `validate` | `register_from_manifest` | Agree |
|---|---|---|---|---|
| key / unset / invalid | collision | warns | raises `Duplicate tool keys detected … crypto_quotes` | ✅ |
| key / unset / invalid | duplicate | warns | raises `… profile` | ✅ |
| key / unset / invalid | clean | silent | registers both | ✅ |
| spec / `" SPEC "` | collision | silent | registers both, no duplicate | ✅ |
| spec | clean | silent | registers both | ✅ |
| **spec** | **duplicate** | **warns** | **does not raise** — registers `company.profile` twice | ⚠️ #162 |

One residual disagreement, and it is exactly the loader gap filed as #162. Two things make it acceptable to ship:

- **It errs in the safe direction.** `validate` warns about something the loader tolerates, not the reverse. A user who heeds the warning is never wrong. The leniency is the loader's, and #149's whole concern was the opposite sign — a validator blessing what the loader refuses. That case is now empty.
- **The wording survives it.** After this round, the duplicate message says "these are the same tool … drop all but one" and no longer asserts registration refuses it. So even in the disagreeing cell the message is not false; it describes a hazard the loader currently tolerates. That is precisely why the "under either name style" claim was removed rather than kept and footnoted.

Still open inside #162, not resolvable here: whether the real FastMCP rejects a repeated tool name or silently keeps the last registration. The `mcp` extra is not installed in the verification environment, so a stub server cannot answer it.

Also confirmed in this pass: invalid and whitespace/case-padded style values (`nonsense`, `" SPEC "`) resolve identically in `validate` and the loader — not by coincidence but because `validate` calls `tool_loader._get_tool_name_style()` rather than reading the variable itself, so there is one source of truth for the style as there now is for resolution. Ten clean-manifest runs across five style values produced zero stderr bytes: no false positives.

Issue #153, opened 2026-08-08
Endpoint.arg_model is a dead mechanism: assigned ~60 times, read never


MehdiZare:
Found while investigating #141.

## What

`Endpoint.arg_model` (`fmp_data/models.py:253`) is declared, assigned roughly 60 times across six domains, and **read by nothing**:

```bash
grep -rn "\.arg_model" --include="*.py" .   # excluding assignments -> no matches
grep -rn "arg_model"    --include="*.py" .  # only the field declaration,
                                            # ~60 `arg_model=` assignments,
                                            # and 4 `arg_model=None` in tests
```

LangChain argument schemas are built **dynamically** instead, in `fmp_data/lc/vector_store.py`:

```python
create_model(
    f"{semantics.method_name}Args",
    **ToolFactory.create_parameter_fields(
        endpoint.mandatory_params, endpoint.optional_params, semantics.parameter_hints
    ),
    ...
)
```

So the `Args` classes in the domain `schema.py` modules never reach a tool, whether or not their endpoint assigns `arg_model`.

## Why this matters

It makes a real distinction look like a bug and a bug look like a distinction.

#141 reports that `investment`, `institutional` and `intelligence` ship `schema.py` modules nothing imports, while six other domains "do wire theirs in". That framing is what the code suggests — but the wiring is inert. All nine `schema.py` modules are equally disconnected from runtime behaviour. There is no anomaly to fix in the three, and "wire them like the others" would assign a field nobody reads.

Anyone reading `arg_model=ETFHoldingsArgs` in an endpoint definition will reasonably assume the model validates that endpoint's arguments. It does not.

## Options

1. **Remove `arg_model` and the nine `schema.py` modules.** Honest, and deletes a lot of dead code — but it removes ~31+ public classes from a published library, so it is a breaking change needing a deprecation cycle (2.7 deprecate, 3.0 remove would match the idiom established in #136/#147).
2. **Remove the field, keep the modules.** The classes stay importable for downstream users; only the inert wiring goes. Smaller break, but leaves nine modules with no consumer.
3. **Make it real** — have the LangChain layer prefer `endpoint.arg_model` when present and fall back to the dynamic builder. Turns dead code into the feature it appears to be, but the hand-written models must then agree with their endpoints, and at least one does not today (see #143: `ETFHoldingsArgs.date` is required while `ETF_HOLDINGS.date` is optional).

[...]

Issue #152, opened 2026-08-08
mutual-fund-holdings returns 404; etf/holdings serves mutual fund symbols


MehdiZare:
Found while probing the live API to settle #143.

## What

`MUTUAL_FUND_HOLDINGS` (`fmp_data/investment/endpoints.py`, `path="mutual-fund-holdings"`) returns **404 for every request**, with or without `date`, for every symbol tried:

```
GET /stable/mutual-fund-holdings?symbol=VFIAX              -> 404 []
GET /stable/mutual-fund-holdings?symbol=VFIAX&date=2024-09-30 -> 404 []
GET /stable/mutual-fund-holdings?symbol=VTSAX             -> 404 []
```

Neighbouring paths are also dead: `funds/holdings`, `mutual-fund/holdings`, `fund-holdings` all 404.

## The working path

`etf/holdings` serves **mutual fund** symbols:

```
GET /stable/etf/holdings?symbol=VFIAX -> 200, 509 rows
```

So FMP appears to have consolidated both product types onto `etf/holdings`, and the separate mutual-fund path is gone.

Probed 2026-08-08 against `stable`.

## Why it matters

`get_mutual_fund_holdings` is broken for every caller — sync, async, LangChain tool and MCP tool alike. It is not a degraded result, it is a 404. Per CLAUDE.md's own guidance ("a 404 indicates incorrect path or deprecated API"), this needs resolving rather than carrying.

It also explains part of #143: the two holdings endpoints disagree about whether `date` is mandatory partly because only one of them has been exercised against a live API recently. `ETF_HOLDINGS.date` is confirmed genuinely **optional** — `etf/holdings?symbol=SPY` returns 505 rows with or without it.

## Options

1. **Point `MUTUAL_FUND_HOLDINGS` at `etf/holdings`.** Smallest fix and matches observed behaviour, but leaves two endpoint objects with identical paths — the shape #130 just finished removing elsewhere.
2. **Deprecate `get_mutual_fund_holdings`** and document `get_etf_holdings` as covering both, following the deprecation idiom now established (`EndpointSemantics.deprecated` from #137, `DeprecationWarning` from #147).
3. Keep it and let it 404, which is the current behaviour and clearly wrong.

Option 2 looks right: one path, one endpoint, and a migration note — but it is a public API change and worth a deliberate decision.

## Note

[...]

PR #154, opened 2026-08-08
fix: remove latent datetime shadowing, guard arg models against their endpoints (#142, #143, #141)

### LINKED ISSUE #141: Three domain schema.py modules are unreferenced by the package (31 arg models)
## What

`investment`, `institutional` and `intelligence` each ship a `schema.py` defining LangChain argument models that **nothing in `fmp_data/` imports**:

| domain | classes in `schema.py` | modules importing it |
|---|---|---|
| `investment` | 10 | 0 |
| `institutional` | 13 | 0 |
| `intelligence` | 8 | 0 |

Six other domains (`alternative`, `company`, `economics`, `fundamental`, `market`, `technical`) *do* wire their schema into `endpoints.py`. These three don't, and no other module picks them up.

Reproduce:
```bash
grep -rl "investment.schema import\|institutional.schema import\|intelligence.schema import" fmp_data/
# (no output)
```

## Why it matters

This is how #139 survived. `fmp_data/investment/schema.py` was completely un-importable under pydantic 2.13 — `ETFHoldingsArgs` declared `date: date`, shadowing the `datetime` import — and **nothing noticed**, because no import path in the package reaches it. It was only found while widening an unrelated drift guard in #138.

Two consequences, both still live for the other two modules:

1. **31 arg-model classes are dead code as far as the package is concerned.** If they are meant to feed LangChain tool registration, that wiring is missing and those endpoints have no argument schemas.
2. **Anything that rots in them rots silently.** `tests/unit/test_imports.py` (added in #140) now catches the un-importable case, but not semantic drift — a wrong field name or type in an unreferenced module has no consumer to break.

## Suggested direction

Decide per domain, rather than assuming one answer:

- If the arg models are **intended** to be part of the LangChain surface → wire them into the domain's `endpoints.py` like the six domains that do, and add coverage that exercises them.
- If they are **superseded/vestigial** → delete them, so nobody maintains schemas that nothing reads.

Worth checking whether the six wired domains and these three differ for a deliberate reason before changing anything.

## Out of scope for #140

#140 fixed the import break and added the guard that would have caught it. Wiring or deleting 31 model classes touches production endpoint definitions and is a separate, deliberate change.

Found during review of #140.

### LINKED ISSUE #142: 23 latent date-field/type shadowing sites can reproduce #139 on any annotation change
## What

#139 was caused by a pydantic field whose *name* shadowed the type imported for its *annotation*:

```python
from datetime import date

class ETFHoldingsArgs(SymbolArg):
    date: date = Field(...)   # PydanticUserError: field name clashes with type annotation
```

An AST scan of `fmp_data/` finds **23 more class attributes named `date` in modules that do `from datetime import date`**. None are broken today, because every one is annotated `datetime` (or `str`) rather than `date` — a *different* name, so pydantic resolves it fine:

```
fmp_data/alternative/models.py:145    IntradayPrice.date: datetime
fmp_data/company/models.py:488        HistoricalPrice.date: datetime
fmp_data/company/models.py:540        IntradayPrice.date: datetime
fmp_data/company/models.py:597        RevenueSegmentItem.date: str
fmp_data/company/models.py:757        AnalystEstimate.date: datetime | None
fmp_data/company/models.py:888        AnalystRecommendation.date: datetime
fmp_data/economics/models.py:85       CommitmentOfTradersReport.date: datetime | None
fmp_data/economics/models.py:468      CommitmentOfTradersAnalysis.date: datetime | None
fmp_data/intelligence/models.py       (13 sites: FMPArticle, PressRelease, ESGData, RatingsSnapshot, StockGrade, ...)
fmp_data/models.py:391                ShareFloat.date: datetime | None
fmp_data/models.py:412                MarketCapitalization.date: datetime | None
```

## Why it matters

Each is a landmine, not a defect. Narrowing any one of them to the semantically-correct type — `AnalystEstimate.date: datetime` → `date`, an entirely reasonable refactor — instantly makes that whole module un-importable, exactly as in #139. The failure is total (the module, not the field) and the error message points at "field name clashing with a type annotation" rather than at the edit that caused it.

The blast radius is wider than #139's: unlike `investment/schema.py`, these are `models.py` modules that the package actually imports, so the break would be immediate rather than dormant.

## Mitigation already in place

`tests/unit/test_imports.py` (#140) imports every module in the package and fails on any exception, so this would now be caught in CI on the commit that introduces it rather than in production. **This issue is therefore hygiene, not an active bug** — it removes the trap instead of relying on the guard to catch someone falling into it.

## Suggested fix

Apply the convention already used in `fmp_data/technical/schema.py`, `fmp_data/investment/schema.py` and `fmp_data/investment/models.py`:

```python
from datetime import date as dt_date
```

Mechanical and low-risk: it changes no field names, so the wire format is untouched. Alternatively `import datetime` and annotate `datetime.date` / `datetime.datetime`, which removes the whole class of collision but is a larger diff.

Reproduce the scan with an AST walk over `fmp_data/` comparing annotated class-attribute names against module-level import names.

Found during review of #140.

### LINKED ISSUE #143: etf_holdings and mutual_fund_holdings disagree on the date parameter (optionality and ParamType)
## What

The two holdings endpoints declare the same logical `date` parameter three different ways, and the argument models disagree with both:

```python
>>> from fmp_data.investment.endpoints import ETF_HOLDINGS, MUTUAL_FUND_HOLDINGS
ETF_HOLDINGS          optional_params:  date  type=ParamType.DATE
MUTUAL_FUND_HOLDINGS  mandatory_params: date  type=ParamType.STRING
```

```python
>>> from fmp_data.investment.schema import ETFHoldingsArgs, MutualFundHoldingsArgs
# both declare `date` REQUIRED, annotated datetime.date
```

So:

| | optionality | type |
|---|---|---|
| `ETF_HOLDINGS` endpoint | optional | `ParamType.DATE` |
| `MUTUAL_FUND_HOLDINGS` endpoint | **mandatory** | **`ParamType.STRING`** |
| both arg models | **required** | `datetime.date` |

Three inconsistencies for one parameter: the two endpoints disagree with each other on optionality, they disagree on `ParamType`, and the arg models are stricter than `ETF_HOLDINGS`.

## Why it matters

`ParamType.STRING` for a date is the same drift class as #131/#134 — a `STRING` param skips date validation and coercion, so a malformed date reaches the API as-is instead of failing locally with a useful message. And an arg model stricter than its endpoint means a caller satisfying the endpoint contract can still be rejected by the tool schema.

This is latent today only because `fmp_data/investment/schema.py` has no consumer in the package (#141). It becomes live the moment those models are wired in.

## What needs deciding

Not obviously a mechanical fix — someone should check the FMP API docs for which endpoint actually requires the date:

1. Is `date` genuinely mandatory for mutual fund holdings but optional for ETF holdings, or is one of them simply wrong?
2. `MUTUAL_FUND_HOLDINGS.date` should almost certainly be `ParamType.DATE` regardless — worth confirming nothing depends on the string passthrough.
3. Once settled, align the arg models with the endpoints and add a guard asserting arg-model requiredness matches endpoint mandatory/optional placement — the drift had no test to catch it.

## Note

`tests/unit/test_investment.py` (added in #140) deliberately does **not** assert requiredness on these fields, precisely to avoid cementing one side of this contradiction. It has a comment pointing here.

Found during review of #140.

MehdiZare:
Closes #142
Closes #143
Closes #141

## #142 — 23 latent sites that could reproduce #139

#139 was a pydantic field whose *name* shadowed the type imported for its *annotation* (`date: date` after `from datetime import date`), which made `fmp_data/investment/schema.py` un-importable and every arg model in it unreachable at runtime.

An AST scan found **23 more class attributes named `date` in modules that import `date` from `datetime`**. None were broken, because each was annotated `datetime` or `str` — a *different* name. But changing any one of those annotations to `date` reproduces #139 exactly.

Rather than fixing 23 instances and leaving the trap open, this makes the shadowing impossible to reintroduce: a guard scans the package by AST and fails on any class attribute whose name matches a name imported from `datetime` in that module, naming file:line. The 23 sites are cleared using the `from datetime import date as dt_date` idiom already used in `fmp_data/technical/schema.py`.

## #143 — the holdings endpoints disagreed about `date`

```
ETF_HOLDINGS          optional_params:  date  ParamType.DATE
MUTUAL_FUND_HOLDINGS  mandatory_params: date  ParamType.STRING
both arg models:      date REQUIRED
```

Settled against the live API on 2026-08-08 rather than by inspection:

- `etf/holdings?symbol=SPY` returns **505 rows with or without `date`** — so `ETF_HOLDINGS.date` is genuinely optional and `ParamType.DATE` is right. `ETFHoldingsArgs` no longer requires it.
- `mutual-fund-holdings` returns **404 for every request**, every symbol, with and without `date`. Its declaration cannot be validated, so the path is left alone and a comment points at **#152**, filed for that decision.

## #141 — the premise was wrong, so the fix is different

#141 reports three `schema.py` modules that nothing imports, while six others "are wired in" via `endpoint.arg_model`.

**`Endpoint.arg_model` has zero read sites.** It is declared once and assigned ~60 times; nothing ever reads it. LangChain builds argument schemas dynamically in `fmp_data/lc/vector_store.py` from `mandatory_params`/`optional_params`/`parameter_hints`. So all nine `schema.py` modules are equally inert — the three are not anomalous, and "wire them like the others" would assign a field nobody reads.

The structural decision (remove the mechanism, remove the field, or make it real) is filed as **#153**.

What #141 correctly identifies is that **nothing catches semantic rot** in these models. `tests/unit/test_imports.py` (from #140) already catches the un-importable case — verified, it walks every module regardless of references — but not drift. So this PR adds the missing guard: **every `Args` model must agree with its endpoint** on parameter names and required-ness. It covers all nine domains, and it is what caught `ETFHoldingsArgs.date` being required while its endpoint's is optional, so #143's fix lands with a regression guard attached.

## Verification

Run in **both** environments, because CI installs no extras and a previous PR in this series passed locally while failing all seven matrix jobs:

```
core-only venv:  1026 passed,  5 skipped
all extras:      1168 passed,  4 skipped
make lint:       clean
```

Both new guards are package-walkers and catch `BaseException`, not `Exception` — `fmp_data/mcp/__main__.py` calls `sys.exit(1)` without the mcp extra, and `SystemExit` is not an `Exception`.

## Follow-ups filed

- **#152** — `mutual-fund-holdings` 404s for every request; `etf/holdings` serves mutual fund symbols
- **#153** — `Endpoint.arg_model` is a dead mechanism across all nine domains


MehdiZare:
@coderabbitai review

MehdiZare:
## Review pass — rebased, verified, all checks green

Rebased onto `dev` (it was `CONFLICTING`/`DIRTY`). The only conflict was `CHANGELOG.md`; the resolution is **purely additive — 13 insertions, 0 deletions** — and keeps dev's corrected "**215** endpoints" in the #129 entry rather than this branch's stale 216. The previously unpushed guard-expansion commit is now included.

### The central claim, verified rather than asserted

#142 claims the `dt_date` aliasing changed no field names, aliases, annotations or wire behaviour. I checked that directly instead of taking it on faith: dumped `model_json_schema()` for every pydantic model in the package on `origin/dev` and on this branch and diffed them.

```
models on dev: 359   models on branch: 359
schemas that differ: 1
   fmp_data.investment.schema.ETFHoldingsArgs
```

**358 of 359 are byte-identical.** The single difference is the one this PR intends:

```
required: ['symbol', 'date']  ->  ['symbol']
date: {"format":"date","type":"string"}
  ->  {"anyOf":[{"format":"date","type":"string"},{"type":"null"}], "default": null}
```

That is #143's fix and nothing else.

### Guards are non-vacuous

Mutation-tested independently, each failing with a precise message:

| mutation | result |
|---|---|
| reintroduce the #139 shape (`from datetime import date` beside an attribute named `date`) | `test_no_class_attribute_shadows_a_datetime_import` fails |
| make `ETFHoldingsArgs.date` required again | rule 1 fails: ``ETFHoldingsArgs.date is required but `etf_holdings` declares it optional`` |
| add a `_KNOWN_PARAM_DRIFT` entry that is not drifting | staleness check fails |

Worth recording that my first attempt at the middle one — changing the annotation from `dt_date | None` to `dt_date` while leaving `Field(None, ...)` — did **not** fail the guard, and correctly so: the field was still optional. The guard keys on `is_required()`, not on the annotation, which is the right thing to key on.

### Test results

```
all extras:  1227 passed,  4 skipped
core-only:   1042 passed, 11 skipped
make lint:   clean
```

Both environments, because CI installs no extras.

### Follow-ups filed rather than widening this PR

- **#167** — the nine `_KNOWN_PARAM_DRIFT` entries now need somewhere to be worked off. `TechnicalIndicatorArgs` is the serious one: `periodLength` and `timeframe` are both mandatory and nothing in the model can fill them, so if these models were ever wired in as #141 originally assumed, every technical-indicator tool would emit rejected requests.

### Reviewer note

Two reviewer subagents were dispatched and neither delivered a report. The verification above is my own. CodeRabbit was triggered manually (auto-review is off for `dev`-base PRs) and replied **"Review rate limited"**, so there is no bot review on this PR either — treat the evidence above as the review, not as a supplement to one.

PR #155, opened 2026-08-08
fix: reconcile required flags with param lists, guard the API docs (#144, #146)

### LINKED ISSUE #144: EndpointParam.required is dead metadata for 13 optional params

### LINKED ISSUE #146: Numbers in docs/api/endpoints.md and docs/mcp/configurations.md are unguarded

MehdiZare:
Closes #144
Closes #146

Both were found during review of earlier PRs in this series.

## #144 — `required` contradicted the list its parameter was declared in

An endpoint declares parameters in two lists, `mandatory_params` and `optional_params`, and each param separately carries a `required` flag. The two say the same thing, so they can disagree — and **14 did**, all sitting in `optional_params` while declaring `required=True`:

```
alternative.CRYPTO_HISTORICAL       start_date, end_date
alternative.FOREX_HISTORICAL        start_date, end_date
intelligence.EARNINGS_CALENDAR      start_date, end_date
intelligence.EARNINGS_CONFIRMED     start_date, end_date
intelligence.DIVIDENDS_CALENDAR     start_date, end_date
intelligence.STOCK_SPLITS_CALENDAR  start_date, end_date
intelligence.STOCK_SYMBOL_NEWS      symbol
intelligence.HISTORICAL_SOCIAL_SENTIMENT  page
```

(The review that filed this counted 13; the walk finds 14.)

**Mostly inert, but not entirely.** `validate_params` skips a `None` value for any param not in `mandatory_params` *before* `validate_value` can consult `required`, so omitting one always worked:

```python
CRYPTO_HISTORICAL.validate_params({"symbol": "BTCUSD"})   # -> {'symbol': 'BTCUSD'}, no error
```

But `validate_value` is public, and calling it directly with `None` raised for a parameter the endpoint treats as optional:

```python
param.validate_value(None)   # ValidationError: Missing required parameter: start_date
```

So one of the two statements was a lie and a reader had no way to tell which.

**Checked against the live API rather than assumed** — the suspicious one was `STOCK_SYMBOL_NEWS.symbol`, since an endpoint named "symbol news" plausibly requires a symbol. It does not: `news/stock` returns 20 rows with *and* without one. All 14 are genuinely optional, so the flags now follow list membership — which is what every consumer actually reads.

A guard asserts the two representations agree in both directions, so they cannot drift apart again.

**Not fixed here:** removing `EndpointParam.required` entirely and deriving requiredness from list membership. That is the honest end state — the flag reaches only one public method — but it touches every endpoint definition and deserves its own change.

## #146 — the API docs had no guard, and had already drifted

`docs/mcp/tools.md` has been guarded by `test_docs_tools_sync.py` for a while, and that guard is why #130's catalog change could not leave a stale number behind. `docs/api/endpoints.md` and `docs/mcp/configurations.md` had no equivalent.

They had already drifted: **the table of contents claimed 47 Market Intelligence endpoints when both the code and the document's own table had 46.** Corrected, and now guarded:

- per-client `(N endpoints)` counts in the TOC asserted against the real `Endpoint` count
- per-section table row counts asserted too, so a row added without touching the heading fails
- every `examples/...` path `configurations.md` names must exist — the check that would have caught it pointing at `examples/mcp_configurations/` when the real directory is `examples/mcp/configurations/`

## Verification

Run in both environments, since CI installs no extras:

```
core-only venv:  1026 passed,  5 skipped
all extras:      1168 passed,  4 skipped
make lint:       clean
```

Both guards mutation-tested — a wrong count and a non-existent path each fail with a message naming the specific mismatch.

Package walks catch `BaseException`, not `Exception` (`fmp_data/mcp/__main__.py` raises `SystemExit` without its extra), and record what they skipped rather than discarding it, so a failure can say what fell out of the scan.


MehdiZare:
@coderabbitai review

MehdiZare:
## Review pass — scope corrected, guard completed, all 17 checks green

### Scope: 4 unrelated commits removed before pushing

The branch was carrying four unpushed commits — an endpoint-deprecation sweep, e2e live-signature probes and endpoint-signature corrections, ~1800 lines across seven client modules. That is a different change with different risk from #144/#146, and pushing it would have taken this PR from ~230 lines to ~2000.

Preserved on `wip/endpoint-deprecation-sweep` and filed as **#168** with notes on what it needs before it can open (rebase, and probably a three-way split, since the deprecation half needs its live-API claims re-probed).

### The fix for #146 had left the document contradicting itself

The TOC was corrected 47 → 46. But `## Market Intelligence` carries its **own** `### 47 endpoints` heading — a third independent copy of that number — and the guard checked the TOC and row *counts*, never the headings. The document still claimed 47 in the section the issue was about.

### The guard promised more than it did

Its module docstring already said *"Row-set equality, not just counts, because a rename miscounts as zero."* The implementation only counted. #146's own suggested fix asked for row-set equality explicitly, for exactly this reason. Implementing it found three rows naming endpoints that do not exist:

| documented as | actually |
|---|---|
| `search_name` | `search-name` |
| `crypto_symbol_news` | `crypto_news_symbol` |
| `forex_symbol_news` | `forex_news_symbol` |

Every count in both sections matched, so no count-based guard could ever have seen these.

### The path column was the one a reader could act on

Comparing paths too found **four rows documenting the bare `/stable/historical-price-eod`** — the exact path `CLAUDE.md` forbids, because these endpoints need a variant suffix. The code correctly requests `historical-price-eod/full`. Anyone following the document rather than the library got a URL the library never sends. `search-name`'s path was documented as `/stable/search_name`, which does not resolve either.

### Every correction landed in the document, not the code

`Endpoint.name` is not cosmetic — it is the cache-key prefix and the argument to `_get_cache_ttl` (`fmp_data/base.py:405-457`). Renaming `search-name` to match its documentation would have silently changed the cache namespace and invalidated any TTL config keyed on the old string. Filed as **#166** instead.

`SKIPPED_MODULES` was populated and never read, so a mismatch found by a partial scan read identically to one found by a complete scan. It is now named in every failure message.

### Mutation-tested, five for five

| mutation | test that catches it |
|---|---|
| wrong TOC count | `test_endpoint_doc_counts_match_the_code` |
| wrong `### N endpoints` heading | `test_endpoint_doc_section_headings_match_the_code` |
| renamed row | `test_endpoint_doc_tables_match_the_code_row_for_row` |
| changed path | same |
| deleted row | same |

Row parsing skips fenced code blocks and carries a ≥250-row floor so it cannot pass by matching nothing.

### On #144

The 14 `required=True` → `False` flips are **behaviour-neutral**, confirmed by reading both paths: `validate_params` does `if value is None and param.name not in mandatory_names: continue` *before* `validate_value` can consult `required`, so the flag was unreachable for every parameter changed.

Filed **#165** for the option #144 itself calls "the honest one" — deriving requiredness from list membership and deleting the field. 546 `required=` assignments, one read site. Too broad for this PR, and it would make the new consistency guard redundant rather than merely satisfied.

### Test results

```
all extras:  1229 passed,  4 skipped
core-only:   1044 passed, 11 skipped
make lint:   clean
```

### Reviewer note

Two reviewer subagents were dispatched and neither delivered a report. The findings above are my own. CodeRabbit was triggered manually (auto-review is off for `dev`-base PRs) and replied **"Review rate limited"**, so there is no bot review here either.

MehdiZare:
Rebased onto `dev` now that #154 has merged (`ccb0a92`). The CHANGELOG merged cleanly — **10 insertions, 0 deletions**, no duplicated entries, #154's entries carried in from `dev` untouched.

Re-verified on the new base, which now also runs #154's `test_arg_model_consistency.py` and `test_datetime_shadowing.py` alongside these guards:

```
all extras:  1234 passed,  4 skipped
core-only:   1049 passed, 11 skipped
make lint:   clean
```

Auto-merge is armed (squash). Note it will not fire on its own: the `Protect Dev` ruleset requires signed commits and these are unsigned, so the PR will sit at `BLOCKED` until it is merged with the usual admin bypass.

MehdiZare:
## Second review pass — three vacuity holes closed, two of them mine

A reviewer pass on the previous commit found three ways the guards could pass while the document was wrong. Each was reproduced before fixing and each now has a mutation test.

**1. The guards only ever proved doc ⊆ code.** Every assertion iterates over the document's *sections*, so a client with an `endpoints` module and no section has nothing to iterate over. I confirmed it: added an undocumented `fmp_data/newthing/endpoints.py` and all six guards stayed green. That is precisely the drift class #146 is about. Now checked in the code → doc direction too.

One half of that finding did **not** reproduce, and I'm recording it rather than quietly acting on it: renaming `## Index` → `## Indexes` was reported as silently dropping a client, but it **fails loudly** — the TOC label stops resolving to a module and `test_endpoint_doc_counts_match_the_code` reports `Indexes: no fmp_data.indexes.endpoints module`. Only the new-client hole was real.

**2. The row-set rewrite lost duplicate detection — a regression I introduced.** `dict(_doc_rows(section))` collapses a row listed twice before anything counts it, so set equality passes and the row count never moves. The `body_rows != counts[client]` assertion I replaced *would* have caught it (7 ≠ 6). Rows are now counted as a list, with duplicates reported by name.

**3. `test_configurations_doc_paths_exist` validated 1 of the 4 paths in the document.** The regex required backticks, and only the *directory* is backticked. The three that matter — `trading_manifest.py`, `research_manifest.py`, `minimal_manifest.py` in `export FMP_MCP_MANIFEST=...` and `create_app(...)` snippets — sit inside fenced blocks with no backticks. So the guard checked the one path nobody copy-pastes and none of the ones they do, while `assert checked > 0` passed and it read as thorough. Matched bare now; floor raised to `>= 4`. All four exist, so it went green immediately.

### Also corrected — things this PR made false or left stale

- **`fmp_data/lc/vector_store.py`** — `create_parameter_fields`' docstring still justified ignoring `EndpointParam.required` with *"13 params sit in `optional_params` with `required=True`"*. As of #144 that is **0** (and it was 14, not 13). This is the exact text a maintainer reads when deciding whether the flag can be trusted, so it now says which half is fixed and held at zero, and which half is still live.
- **`CHANGELOG.md`** — `## Unreleased` had two `### Fixed` blocks. The second one was added by this PR. Merged.
- **The `BaseException` rationale** cited `mcp/__main__.py`, which the `.endpoints` filter excludes and which can never reach that import. More usefully: the failure it gestures at is raised by `walk_packages` from the `for` header, *outside* the `try` — so an `onerror` callback was needed as well. Both now record what fell out.
- Dropped the no-op `_LABEL_TO_CLIENT["sec"] = "sec"`.

### Verification

| mutation | caught by |
|---|---|
| duplicated doc row | `test_endpoint_doc_tables_match_the_code_row_for_row` |
| undocumented new client | `test_every_client_has_a_documented_section` |
| manifest path that does not exist | `test_configurations_doc_paths_exist` |

plus the five from the previous commit, all still passing.

```
all extras:  1235 passed,  4 skipped
core-only:   1050 passed, 11 skipped
make lint:   clean
```

### Deferred, deliberately

Three items are real but belong elsewhere, and are filed rather than folded in: `ParameterHint.required` is a *third* declaration of the same optional/mandatory fact with no consumer at all (the #144 species again); the mirror contradiction of 13 params in `mandatory_params` carrying a `default`; and the fact that three separate `pkgutil` endpoint walks now exist across the test suite, which wants one `conftest.py` fixture rather than a fourth copy. Tracked on **#165**.

PR #169, opened 2026-08-09
fix: retire 28 endpoints FMP no longer serves, correct 6 signatures, add a live signature suite

### LINKED ISSUE #152: mutual-fund-holdings returns 404; etf/holdings serves mutual fund symbols

MehdiZare:
Closes #152

~~Stacked on **#155**~~ — #155 is merged. This branch has been rebased onto `dev` and now contains only its own 11 commits.

## Where this came from

The user asked for a full end-to-end check against the live API "so the API signature matches our expectation and implementation". All 275 endpoint declarations were probed. The results were worse than the single endpoint #152 reported:

| | |
|---|---|
| Answered with data | 194 |
| **Dead paths (404)** | **28** |
| **Signature bugs** | **6** |
| CSV batch endpoints (not faults) | 19 |
| Empty due to sample values | 23 |

## The 28 dead paths: the library already shipped every replacement

The plan was to repoint them. That was wrong, and finding out why is the substance of this PR.

**Every one of the 15 repoint targets already exists as a live endpoint with a working client method.** `alternative.CRYPTO_QUOTES` declares the dead `quotes/crypto` while `batch.BATCH_CRYPTO_QUOTES` has served `batch-crypto-quotes` all along; `company.SHARE_FLOAT` already serves `shares-float`; `investment.ETF_HOLDINGS` already serves `etf/holdings`. These were not endpoints needing new paths — they were **dead declarations left behind after a migration that already happened**. Repointing would have created 15 duplicate `(path, params)` pairs, the exact duplication #130 and #136 removed.

Independently, **9 of the 15 would have failed shape validation** — `batch-*-quotes` omits `timestamp`, `ratings-historical` omits three required fields, `funds/disclosure-holders-search` returns entity-registry data entirely unlike the holder models. A clean 404 would have become a runtime parse failure.

So all 21 are **deprecated with a pointer to the method that already works**, not repointed. 7 more have no replacement at all and are marked withdrawn.

## 6 signature bugs the API reported directly

| endpoint | was | now |
|---|---|---|
| `market.CIK_SEARCH` | `ParamType.STRING` sent an unpadded CIK | `ParamType.CIK` |
| `sec.SEC_FILINGS_SEARCH_{SYMBOL,CIK,FORM_TYPE}` | `from`/`to` optional | mandatory — the API 400s without them |
| `alternative.{CRYPTO,COMMODITY}_INTRADAY` | `symbol` as a path segment | query param |
| `investment.MUTUAL_FUND_HOLDINGS` | `symbol` declared `location=PATH` on a path with no `{symbol}` | query param — it was **silently dropped** before |
| `sec.INDUSTRY_CLASSIFICATION_SEARCH` | no mandatory params | one-of constraint documented |

The `MUTUAL_FUND_HOLDINGS` one was found during implementation, not in the original audit — the caller's symbol went nowhere.

## New: `tests/e2e/test_live_signatures.py`

Opt-in on `FMP_TEST_API_KEY`, so the extras-free CI matrix is unaffected. Asserts no endpoint 404s, mandatory params are really mandatory, optional really optional, path templates are satisfiable, and — the reverse check — that every endpoint exempted for being deprecated **really is still dead**, so a revived endpoint is noticed too.

**Result: 6/6 passing. 404s: 0, down from 28.**

Two mistakes this file made before it worked, both recorded in its docstring so they are not reintroduced:

- **Building URLs by hand.** It now probes through `validate_params` → `build_url` → `get_query_params`, so aliased params and `{placeholder}` paths are exercised as a real caller would. The hand-built version reported false failures on both.
- **Inferring "optional" from a row count.** The API defaults to AAPL — our own sample value — so `key-executives` with and without `symbol` returns byte-identical Apple data. That produced 14 false positives. It now compares against a non-default sentinel, and `SENTINELS` is deliberately narrower than `SAMPLES` because `query` and `name` are polysemous across endpoints and an invalid sentinel triggers the same fallback.

## `@deprecated`, not `@removed`

All 17 affected endpoints carry `allow_empty_on_404=True`, so they already return `[]` today. `@removed` raises, which is a breaking change in a minor release. This follows the 2.6-deprecate / 3.0-remove cycle from #136/#147.

## Also here

`generate_manifest` now reports three exclusion categories rather than one — withdrawn, deprecated, collision loser — each with its reason. That distinction became load-bearing once the `alternative` side of both ambiguous keys was withdrawn: reporting those as collision losers would tell users the wrong thing about why a tool vanished.

## Verification

```
make lint                clean
unit (extras)            1234 passed,  4 skipped
unit (core-only)         1046 passed, 11 skipped
e2e (live API)              6 passed
```

Core-only is run separately because CI installs no extras, and that gap has bitten twice in this series.

---

## Correction: `fmp-mcp generate` was emitting dead tools, and preferring them

This section was missing from the original description. It is the most user-visible bug in the PR, and it was nearly lost.

The last failing test (`test_a_default_yields_to_an_explicit_pick_of_the_other_side`) looked like a stale assertion — the scenario it described no longer existed, so the tempting fix was to drop the assertion and document the path as unreachable. **That would have buried a live bug.** Checking the mechanism instead showed `_startable_catalog` filtered `DEPRECATED_TOOLS` but not `WITHDRAWN_TOOLS`, so nothing was being dropped for withdrawal at all.

Measured on the pre-fix tree vs. the fix:

```
                                   before      after
generated tools                       218        201
withdrawn specs written in             19          0
kept alternative.crypto_quotes   yes (404)         no
kept batch.crypto_quotes    no (excluded!)   yes (200, 4771 rows)
```

Two distinct defects:

1. **All 19 withdrawn specs were written into every generated manifest** — `fmp-mcp generate` produced a server advertising 19 tools that can only answer empty. The same defect removed from `DEFAULT_TOOLS`, still live on the generate path.

2. **The dead spec won the collision tie-break.** `_preferred_spec` breaks ties on `DEFAULT_TOOLS` membership. Once the dead spec left the defaults, *neither* side was a default, so the tie-break fell through to sorted order and put `alternative` first. The generated manifest kept `quotes/crypto` (404) and excluded `batch-crypto-quotes` (200, 4771 rows) — then reported the **working** tool as the one it had dropped:

   ```
   batch.crypto_quotes (kept alternative.crypto_quotes, ...)
   ```

   This was introduced by the `DEFAULT_TOOLS` change in this same PR. The failing test was the only thing pointing at it.

Fix: withdrawn specs are dropped **before** the collision pass, so nothing dead is emitted and no dead spec can outrank a live one. With one side of each ambiguous pair withdrawn, no ambiguity reaches the manifest — which is why the `FMP_MCP_TOOL_NAME_STYLE=spec` guidance is now correctly absent rather than merely missing.

The three drop reasons are reported separately rather than merged, because merging would lie. The deprecation note promises *"the replacement named beside each one is already included below, so no capability is lost"* — untrue for a withdrawal, since six have no successor. Withdrawals get their own section naming the nearest live tool where one exists (`alternative.crypto_quotes` → `batch.crypto_quotes`) and saying plainly *"FMP publishes no replacement"* where none does.

**Reviewer note:** `tests/unit/test_mcp_cli.py` was edited concurrently by two sessions during this work. The pushed version is green in both environments, but it is worth a closer read than the rest of the diff, since neither author wrote all of it.

---

## Post-review: rebase onto `dev`, and the patch-coverage failure

**Rebased.** #155 was squash-merged, so its two commits (`cf0559d`, `82563ea`) were dropped via `git rebase --onto origin/dev 82563ea`. The resulting `CHANGELOG.md` conflict was resolved by folding this branch's seven new bullets into the `### Fixed` section `dev` already has, rather than adding a second `### Fixed` heading under `## Unreleased`. `check-rebase` is green.

**`codecov/patch` was the one failing check** — 75.25% against an 80% target. Every one of the 24 uncovered lines was the body of a method this PR retired. That is not a coincidence and it was not noise: these methods wrap endpoints FMP no longer serves, so nothing had ever called them, and the decorators were therefore unverified. A `@deprecated` applied to the wrong method would have looked identical to a working one.

`tests/unit/test_deprecated_methods_warn.py` closes it by **calling every retired method with arguments bound from its own signature**. That detail is the whole point. A bare `method()` call fires the decorator and then dies on the missing argument *inside the wrapper's `*args` passthrough, before the body runs* — so the warning assertion passes while the stub `return []`, the line that actually encodes the withdrawal, stays uncovered. The first version of this file did exactly that and reached only 80.41%.

With arguments bound, the body executes and the two things that matter can be asserted:

- the method **returns empty** (`[]` or `None`), and
- it **does not reach the client** — a retired endpoint that still issued a request would spend a caller's rate-limit slot to earn a 404.

`@removed` is asserted against its own contract (it raises `RemovedEndpointError`) rather than against the warning, since it shares the `__fmp_deprecated__` stamp and a warning-only check would report those three as broken. Nothing is swallowed: an unexpected exception fails the test and names the method.

```
patch coverage       75.25%  ->  100%
```

## Verification

```
ruff check fmp_data tests          clean
ruff format --check fmp_data tests 211 files already formatted
unit (extras)                      1244 passed,   3 skipped
nox tests-3.12(core)               1068 passed, 289 skipped
e2e without FMP_TEST_API_KEY          6 skipped
```

Core-only is run separately because CI installs no extras, and that gap has bitten this branch series twice.

## Filed rather than fixed here

- **#170** — `removed()` in `fmp_data/helpers.py` has no `iscoroutinefunction` branch, unlike `deprecated()`, so the three `@removed` `async def` methods in `fmp_data/intelligence/async_client.py` are not coroutine functions. Pre-existing: the `helpers.py` diff in this PR only adds the `__fmp_deprecated__` stamps. Out of scope, so it is an issue rather than a wider change here.


MehdiZare:
## Code review round: two Critical findings, both reproduced and fixed

A reviewer pass over the rebased diff returned two Critical findings. Both were reproduced independently before any code was touched — they are real, and both were promises this PR's own CHANGELOG made and did not keep.

### Critical 1 — the withdrawn MCP branch was unreachable

`_warn_if_deprecated` grew a `WITHDRAWN_TOOLS` branch with wording that carefully avoids promising a drop-in. Nothing ever reached it.

`_resolved` set `ResolutionStatus.DEPRECATED` only from `DEPRECATED_TOOLS`, and the two sets are disjoint *by assertion*, so every withdrawn spec resolved as plain `RESOLVED` and the call site's `if resolution.is_deprecated` gate was never true. Measured before the fix:

```
alternative.crypto_quotes       status=RESOLVED  is_deprecated=False  warnings=0
institutional.asset_allocation  status=RESOLVED  is_deprecated=False  warnings=0
```

The unit test passed because it called the private helper directly — a test shape that asserts message text and structurally cannot detect that no caller reaches it.

The same root cause blinded `fmp-mcp validate`: `_classify_manifest_entries` bucketed on `resolution.replacement is not None`, and a withdrawal deliberately carries no replacement, so all 19 fell through every branch and a manifest full of dead tools was reported as perfectly healthy. `validate` and `generate` disagreeing is the drift #149 exists to prevent.

Fixed by making withdrawal a first-class status:

- `ResolutionStatus.WITHDRAWN` is a real member; registration gates on a new `is_retired` (deprecated **or** withdrawn)
- `is_resolved` includes it — a withdrawn tool still registers and answers empty, so explicit manifests keep working until 3.0
- `Resolution.successor` is kept apart from `replacement`, because a deprecation is a rename with a drop-in while a withdrawal is a migration, and 4 of the 19 have no successor at all. Collapsing them would make `validate` print "use X" for an endpoint with no X
- `validate` reports withdrawn as its own category
- The regression test goes **through** `_resolved` and sweeps every entry in `WITHDRAWN_TOOLS`

After: `status=withdrawn is_retired=True warns=1`, with the no-replacement variant correctly saying so.

### Critical 2 — LangChain still indexed all 19 dead endpoints

The MCP surface was taught to drop withdrawn tools. The semantic-search surface was not — no `mapping.py` was touched, and all 19 sat at `deprecated=False`.

The mitigation this PR's CHANGELOG leaned on does not reach the index: `registry.py` builds embedding text from `natural_description`, `example_queries` and `related_terms`. The endpoint `description` — where every "DEPRECATED and non-functional … Do not select it" sentence was added — **is never embedded**. `stock_news_sentiments` is excluded because `semantics.deprecated=True` hard-filters it, not because of its wording.

So a LangChain agent asked for crypto quotes or analyst price targets still retrieved the dead endpoint, called it, and got an empty *success* indistinguishable from "no data matched" — the exact failure #137 exists to prevent.

Fixed by using the mechanism the codebase already has: all 19 `EndpointSemantics` now carry `deprecated=True`, which `_is_deprecated` applies to the add, load and search-result paths at once. The guard in `test_deprecated_endpoint_flags.py` now **derives** its expected set from `WITHDRAWN_TOOLS`, so the next withdrawal cannot fix one surface and forget the other — that cross-surface check is what was missing.

### Also fixed

- `fmp-mcp generate --tools <withdrawn-spec>` warns on stderr. The default path drops withdrawn specs before the collision pass, so `--tools` was the only route by which a dead tool reached a manifest, and the only place one was written with nothing said about it
- `_add_defaults` skips withdrawn as well as deprecated
- The async sweep in `test_deprecated_methods_warn.py` asserts the empty return its sync twin already did — the async wrapper is the path most likely to diverge
- **`make test` no longer fires the live suite.** It sources `.env` and promotes `FMP_API_KEY` to `FMP_TEST_API_KEY`, and `testpaths` is `tests`, so a maintainer with a populated `.env` was firing ~700 live requests against a 250/day quota. `tests/e2e` is marked `live` and deselected by `-m "not live"` in `addopts`; run with `pytest tests/e2e -m live`
- Two CHANGELOG claims were false as written. Rather than soften the text, the code now makes them true, and the entry says plainly which mechanism does the excluding

### Adversarial re-check of the fixes

```
overlap DEPRECATED & WITHDRAWN            none
withdrawn still in DEFAULT_TOOLS          none
withdrawn specs misresolving              none (all 19 -> WITHDRAWN)
deprecated specs misresolving             none
successors that are themselves retired    none
```

### Verification

```
ruff check fmp_data tests          clean
ruff format --check                211 files already formatted
mypy                               clean, 121 source files
unit (extras)                      1247 passed,   3 skipped
nox tests-3.12(core)               1069 passed, 283 skipped, 6 deselected
e2e default / -m live without key  6 deselected / 6 skipped
patch coverage                     100%
```

### Filed rather than fixed here

- **#170** — `removed()` lacks the `iscoroutinefunction` branch `deprecated()` has, so three `@removed` async methods are not coroutine functions. Pre-existing
- **#171** — four deferred minors: e2e exemption keyed by bare attribute name (zero collisions today, latent), no withdrawn counterpart to the replacement-naming assertion, one misnamed test, and reconciling the 28/21/19/8/7/4 counts in the CHANGELOG
- **#172** — LangChain tools dispatch via `client.request`, bypassing client-method defaults. Not a regression, but the same latent gap

PR #173, opened 2026-08-09
fix(lc): harden the hint-duplication and deprecation-flag guards (#157, #159)

### LINKED ISSUE #157: Hint duplication guard is keyed on binding name, so same-concept-different-name copies stay invisible
Split out of the review of #150 (#135's consolidation).

## The gap

`tests/unit/test_hint_consolidation.py` compares `ParameterHint` bindings **by attribute name**. That catches the #135 defect exactly — two modules binding `SYMBOL_HINT` to different contents — but it cannot see a divergent copy that simply picked another name.

Concrete instance, present today:

- `fmp_data/technical/mapping.py:31,41` declare local `FROM_DATE_HINT` / `TO_DATE_HINT`.
- These are the same concept as `fmp_data.lc.hints.DATE_HINTS["start_date"]` / `["end_date"]`, with different contents: examples `2024-01-01`/`2023-12-01` vs `2023-01-01`/`2022-12-31`, and the shared version's bare `(\d{4}-\d{2}-\d{2})` pattern is missing.

Both guards pass: `test_no_two_modules_define_the_same_hint_differently` never compares them (different names), and `test_cross_domain_hints_originate_in_lc_hints` sees a single owner (only `technical` uses them).

This is not a #135 regression — the names are single-owner and were never in the ten consolidated concepts — but the guard reads as "hints cannot silently diverge", and for this shape it does not hold.

## Options

1. **Fold `technical` onto the shared date hints.** Straightforward, but it changes the embedding text and tool descriptions of nine indicator tools, so it wants its own before/after review rather than riding along in a consolidation PR.
2. **Guard by concept, not name** — cluster hints by content similarity (shared patterns, or the parameter name they are bound to across the catalog) and flag near-duplicate clusters that are not sourced from `fmp_data.lc.hints`. Catches the general case, needs a threshold that does not cry wolf.

(2) is the durable fix; (1) is worth doing regardless, since the divergence is real.

### LINKED ISSUE #159: _is_deprecated takes Any and getattr-defaults, so a renamed field silently disables filtering
Split out of the review of #150 (#137). Low severity, raised by two of three reviewers.

`fmp_data/lc/vector_store.py::_is_deprecated` is typed `info: Any` and reads the flag with `getattr(info.semantics, "deprecated", False)`. Rename or drop the field and every filtering site degrades to "nothing is deprecated" — silently, with the suite green except for the pins listed below.

It is deliberate: the unit tests drive `add_endpoint` / `add_endpoints` / `get_tools` through duck-typed stubs, and `EndpointInfo` is only importable with the `langchain` extra. `tests/unit/lc/test_vector_store.py` even needed a `_live_endpoint_info` fixture because `Mock(spec=EndpointInfo)` returns a *truthy* Mock for `.deprecated`, which is evidence of the same fragility from the other direction.

Current mitigations: `test_deprecated_defaults_to_false` pins `EndpointSemantics.model_fields["deprecated"]` (and now runs in the no-extras matrix), and `test_search_filters_deprecated` exercises real semantics through the real registry.

## Fix

Type the parameter `EndpointInfo` and read `info.semantics.deprecated` directly, so mypy verifies the field exists. That means the stub-driven tests need real `EndpointInfo` values — the `real_store` fixture in `tests/unit/lc/test_deprecated_endpoints.py` already shows the shape, since `SearchResult.info` forced it there.

Worth doing when those fixtures are next touched rather than on its own.

MehdiZare:
Closes #157
Closes #159

Both were split out of #150's review. Both are guards that did not catch what they claimed to — the failure mode this repo has been repeatedly bitten by, and the one that is hardest to notice because a passing test looks identical either way.

## #157 — the hint guard enforced "one definition per name", not per concept

`tests/unit/test_hint_consolidation.py` keyed on the *binding name*, so two modules defining the same concept under different names stayed invisible. A `SYMBOL_HINT` in one module and a `TICKER_HINT` in another, same meaning and different extraction patterns, is exactly the #135 defect and passed.

## #159 — a guard whose failure mode was silence

`fmp_data/lc/vector_store.py::_is_deprecated` was typed `info: Any` and read the flag with a `getattr` default. Renaming `EndpointSemantics.deprecated` would have silently disabled the filtering added for #137 rather than failing — the vector store would quietly start serving deprecated endpoints again with every test still green.

## Verification

```
make lint          clean
unit (extras)      1247 passed,  4 skipped
unit (core-only)   1057 passed, 11 skipped
```

Core-only is run separately because CI installs no extras, and that gap has bitten three times in this series.

Both guards are mutation-tested — the defect each exists to catch was introduced, the failure observed, and the tree restored.


MehdiZare:
## Code review response

Reviewed against #157 and #159. Verdict: **merge with fixes** — no Critical. 2 Important, both real and both fixed in `d2f8124`, plus the Minor items.

**#159 checks out end to end.** The type chain is concrete all the way down (`EndpointInfo.semantics: EndpointSemantics`, `deprecated: bool`), and every call site feeds `registry.get_endpoint()` → `EndpointInfo | None`, so mypy genuinely checks them. Renaming the field in a scratch copy produces `error: "EndpointSemantics" has no attribute "deprecated"` plus 5 runtime failures; the old `getattr(..., False)` passed both.

**The #157 guard is core-only safe** — simulated by blocking `langchain*`/`mcp`/`faiss`: 36 candidates still found, guard runs and passes. The repeated failure mode of this series is avoided.

### Important 1 — the floor was pinned to the defect count

This is the sharper of the two. `MIN_SAME_CONCEPT_CANDIDATES = 20` counted *findings*, and all 36 candidates come from the one divergence the PR allowlists. So applying the fix #157 itself calls for — folding `technical/mapping.py`'s date hints onto the shared `DATE_HINTS` — failed this guard:

```
AssertionError: only 0 same-concept candidate pairs found; the comparison
in _same_concept_candidates may not be running
assert 0 >= 20
```

A correct fix, failing with a message that misdiagnoses it as a broken detector. The natural response is to drop the floor to zero, which reinstates exactly the vacuous pass the floor exists to prevent.

Now floors on pairs **compared** (5188 observed, floor 3000) rather than pairs flagged. Verified both directions:

- the sanctioned #157 fix + emptied allowlist → **9/9 pass**
- short-circuiting the detector → fails with `only 0 hint pairs were compared`, which is now an accurate diagnosis

A regression that keeps comparing but stops matching is caught by the stale-allowlist assertion, which was already there — that, not the positives floor, was doing the real work.

### Important 2 — `_is_extension` exempted equality, hiding the likeliest duplicate

`set_a <= set_b or set_b <= set_a` treats *identical* pattern sets as a refinement. The docstring justifies exempting a strict refinement; equality is the degenerate case and isn't one. Confirmed with a new binding carrying patterns byte-identical to `DATE_HINTS["start_date"]` under a different name, with drifted `examples`/`context_clues` — it **passed all 9 tests**. Different name, so #135's name guard is blind; identical patterns, so `_is_extension` short-circuits. That is precisely the #135 defect class this guard advertises catching, and `examples`/`context_clues` are what get rendered into tool descriptions and embedding text.

Now a strict subset (`<`). The verbatim copy is caught, and the reviewed `PERIOD_HINT`/`PERIOD_WITH_FISCAL_HINT` split still does not trip — no false positive.

### Minor items

- **3, 4** — stale constant name in a comment and `(\d{4}-\d{2}-\d{2})` described as 18 chars (it's 19). Both fixed.
- **5** — allowlist `allowed_a or allowed_b` exempts every pair a listed binding takes part in. Kept, now documented: the allowlisted hint is a local copy of a shared definition and pairs with it in every importing module, so requiring both sides would mean enumerating all those pairs. A genuinely new drifted copy still trips, because it also pairs with the un-allowlisted `fmp_data.lc.hints` binding.
- **6** — folding the `technical` date hints was tracked only by a code comment. Now **#179**, referenced from `KNOWN_SAME_CONCEPT_DIVERGENCES`.
- **7** — `_hint_bindings` sees module-level bindings plus one level of dict, so the ~20 inline `ParameterHint(...)` literals inside `parameter_hints={...}` are invisible to both guards. Pre-existing; now stated in the module comment so this file no longer reads as exhaustive, and noted in #179.
- **8** — good catch on the CHANGELOG. "An `AttributeError` at the first call" holds on the `search`/`get_tools` paths; on `add_endpoints`, `_classify_endpoint`'s broad `except Exception` turns it into a skipped endpoint plus an ERROR log. Both are loud — which is the point versus the old silent `getattr` — but the wording now says which is which.

Full suite after the changes: **1247 passed, 4 skipped**; ruff and mypy clean.

PR #174, opened 2026-08-09
deprecate Endpoint.arg_model and the hand-written arg models (#153, #167)

### LINKED ISSUE #153: Endpoint.arg_model is a dead mechanism: assigned ~60 times, read never

### LINKED ISSUE #167: Nine arg-model fields name no parameter of their endpoint (TechnicalIndicatorArgs cannot build a valid request)
Baselined in `_KNOWN_PARAM_DRIFT` by the new guard in #154/#141. The guard freezes the set — it may not grow, and a fixed entry must be deleted — but nothing tracks actually shrinking it. This is that tracking issue.

Related to #153 (`Endpoint.arg_model` is assigned ~60 times and read never), but distinct: #153 is about the mechanism being dead, this is about these specific models being *wrong* if the mechanism is ever revived.

## The worst one: TechnicalIndicatorArgs

Shared by all 9 indicator endpoints. Its field names and the endpoint's parameter names have no overlap and no alias bridging them:

| model field | endpoint parameter |
|---|---|
| `start_date` | `from` |
| `end_date` | `to` |
| `period` | `periodLength` (**mandatory**) |
| `interval` | `timeframe` (**mandatory**) |

Both `periodLength` and `timeframe` are mandatory and nothing in the model can fill them, while `period` and `interval` would never reach the wire. If these models were ever wired into tool generation as #141 originally assumed they were, every technical indicator tool would emit requests the endpoint rejects.

## The rest

| model.field | reason |
|---|---|
| `SearchArgs.page` | `search-name` takes no page parameter |
| `Form13FArgs.filing_date` | endpoint takes `year`+`quarter`, not a date — the model can express neither |
| `AssetAllocationArgs.filing_date` | endpoint parameter is plain `date` |
| `InstitutionalHoldingsArgs.include_current_quarter` | matches no parameter; model also cannot supply the mandatory `year`/`quarter` |
| `FailToDeliverArgs.limit` | `fail_to_deliver` paginates with `page` only |

## Why it is not urgent, and why it should not be forgotten

Nothing reads these models today, so none of this is user-visible — that is exactly what let them rot. The guard added in #154 (`tests/unit/test_arg_model_consistency.py`) means they cannot get worse and cannot be quietly "fixed" without the baseline entry being deleted. But the baseline only shrinks if someone works it.

Settling #153 first would decide whether the right fix is to correct these models or delete them.

MehdiZare:
Closes #153
Closes #167

## The mechanism was never wired up

`Endpoint.arg_model` was declared once, assigned **68 times** across six domains, and read by nothing — `grep -rn "\.arg_model"` finds no read site in the package. LangChain builds argument schemas **dynamically** in `fmp_data/lc/vector_store.py` from each endpoint's `mandatory_params`/`optional_params` plus its `parameter_hints`, so the **106 `*Args` classes** across nine `schema.py` modules never reached a tool either way.

That mattered because the wiring *reads* as a contract. #141 was filed on exactly that misreading — three modules looked anomalous for not assigning a field the other six assigned, when in truth all nine were equally inert.

## The models had already rotted, which is why this is a removal and not a revival

Nothing read them, so nothing could notice them drifting. #167 measured the damage. The worst case, `TechnicalIndicatorArgs`, is shared by all **9** indicator endpoints and has **zero** parameter-name overlap with its endpoint:

| model field | endpoint parameter |
|---|---|
| `start_date` | `from` |
| `end_date` | `to` |
| `period` | `periodLength` (**mandatory**) |
| `interval` | `timeframe` (**mandatory**) |

Both mandatory parameters are unfillable from the model, and `period`/`interval` would never reach the wire. Had these been wired into tool generation — the alternative this PR rejects — every technical indicator tool would have emitted requests the endpoint rejects. Five further models have the same class of defect.

A live API audit this week separately found the endpoint declarations themselves wrong in ~33 places. A second hand-maintained layer would have encoded those same errors somewhere nothing cross-checks. This drift class already produced #135, #144, #146 and #167.

## What is removed now vs deprecated

**Removed now** — internal wiring nothing depends on: the 68 `arg_model=` assignments and the schema imports that existed only to feed them.

**Deprecated, removed at 3.0** — public surface that stays importable: the `arg_model` field, and the arg-model classes themselves. Verified behaviour:

```
setting arg_model=...        -> DeprecationWarning
instantiating ProfileArgs()  -> DeprecationWarning
importing the module         -> silent (an indirect import should not warn)
```

Warning on *use* rather than on import is deliberate — an import-time warning fires for transitive importers who never touch the classes, which is how deprecation warnings get filtered out wholesale.

This follows the 2.7-deprecate / 3.0-remove cycle established in #136/#147, with a version-guard test that fails at the 3.0 bump so the cleanup cannot be forgotten.

## Verification

```
make lint          clean
unit (extras)      1254 passed,  4 skipped
unit (core-only)   1064 passed, 11 skipped
```

Core-only is run separately because CI installs no extras, and that gap has bitten three times in this series.


MehdiZare:
## Code review response

Reviewed against #153 and #167. Verdict: **merge with fixes** — no Critical. 2 Important and 4 Minor, all resolved below in `5a8d556`.

The crux question for this PR — *is this a deprecation, or a removal wearing a deprecation label?* — comes out on the right side. Verified empirically, not read:

```
Endpoint(..., arg_model=X)        -> DeprecationWarning, field value preserved
Endpoint(...) / arg_model=None    -> silent
"arg_model" in ep.model_dump()    -> True   (serialization shape unchanged)
import ETFHoldingsArgs            -> silent
TechnicalIndicatorArgs(symbol=..) -> DeprecationWarning naming the class
```

The `model_validator(mode="before")` shim is the right mechanism — a `@property` would have collided with the field declaration. Coverage was confirmed independently of the PR's own guard by walking every `schema` module: **116 classes, zero misses**. And the guards are not vacuous: five separate mutations were each caught.

### Important, both fixed

**1 — `endpoint.arg_model` now reads `None`, silently, and the CHANGELOG said otherwise.** The claim *"no code path reads the field, so deleting the assignments changes no behaviour"* is true **inside the package** and false for a consumer:

```python
from fmp_data.technical.endpoints import SMA
SMA.arg_model   # was TechnicalIndicatorArgs, now None — zero warnings on read
```

That is the one thing in this deprecation that changes silently, and it was filed under `### Deprecated` while the same CHANGELOG marks smaller changes `**BREAKING:**`. Now scoped to *"inside this package"* and given its own `### Changed` entry naming the behaviour and pointing at `mandatory_params`/`optional_params` as the replacement. No code change — a read-side warning needs a descriptor and isn't worth it for a field that's gone in 3.0.

**2 — `model_json_schema()` didn't warn** (`fmp_data/schema.py`). Schema generation was the gap: an argument model a downstream caller kept is most naturally handed to a tool/function-schema generator, which never validates anything. `DeprecatedArgModel` now also warns from `__get_pydantic_json_schema__`.

Import stays silent — pydantic v2 builds JSON schemas lazily, so nothing fires at class-creation time. Verified under `-W error::DeprecationWarning`, and now pinned by a subprocess guard so a future pydantic that generated eagerly would fail the build rather than turn every `import fmp_data` into noise.

### Minor, all fixed

**3 — real guard hole.** The walk filtered on `"Args"`, so `BaseSearchArg` and `BaseExchangeArg` — public, deprecated, no in-tree subclass — were covered by nothing; reverting either left the file green. Now `endswith(("Arg", "Args"))`, floor `100 → 110`. Confirmed: reverting `BaseSearchArg`'s base now fails two tests.

**4 — two inaccurate docstrings.** *"the four roots outside `BaseArgModel`"* — there are **seven**; the parametrization covers three plus `BaseArgModel`, and exhaustiveness is the walk's job. And the `except BaseException` rationale cited `fmp_data/mcp/__main__.py`, which this walk never imports (it only imports `.schema`/`.endpoints`, and `walk_packages` imports packages, whose `__init__` files are empty here). The catch is defensible as forward-looking; it's now described that way.

**5 — `stacklevel=2` misattribution**, reported at `pydantic/main.py` rather than the caller because construction crosses pydantic-core frames. No correct `stacklevel` exists through a compiled boundary, so this is now an explicit comment rather than a wrong number. Per-class filtering still works — the class name is in the message.

### One correction to the review

The review counted **12** singular `*Arg` classes, for 118 total. The actual figure is **10**, for **116** — measured after widening the filter:

```
total 116 | *Args 106 | *Arg 10
```

The floor and every comment use 116.

### Deferred, with an issue

**6 — the ~100 concrete classes carry no `.. deprecated::` marker**, only the 8 roots do, so `help()`, IDE hover and Sphinx show nothing for `ETFHoldingsArgs` et al. → **#178**. The fix is either 100+ mechanical docstring edits or an `__init_subclass__` that appends the marker; neither belongs in the PR that establishes the mechanism. The runtime warning does fire for these classes, so the gap is docs-only.

### Mutation evidence for the new guards

| mutation | caught by |
|---|---|
| revert `BaseSearchArg` to plain `BaseModel` | `test_every_arg_model_is_marked_deprecated` + a param case |
| delete the `__get_pydantic_json_schema__` warning | `test_generating_a_json_schema_warns` |
| build a JSON schema eagerly at import | `test_importing_an_arg_model_stays_silent` |

Full suite after the changes: **1256 passed, 4 skipped**; ruff and mypy clean.

PR #175, opened 2026-08-09
fix(helpers): @removed must preserve coroutine-ness, like @deprecated (#170)

### LINKED ISSUE #170: @removed does not preserve coroutine-ness, unlike @deprecated
## What

`removed()` in `fmp_data/helpers.py:87-92` returns a plain synchronous wrapper regardless of what it decorates:

```python
def decorator(func: F) -> F:
    @functools.wraps(func)
    def wrapped(*_args: Any, **_kwargs: Any) -> Any:
        raise RemovedEndpointError(func.__name__, reason)

    wrapped.__fmp_deprecated__ = True
    return wrapped
```

Its sibling `deprecated()` (same file, lines 36-52) *does* branch on `inspect.iscoroutinefunction(func)` and returns an `async def` wrapper so that `inspect.iscoroutinefunction` stays True — the docstring calls this out explicitly:

> Coroutine functions are wrapped with an async wrapper so `inspect.iscoroutinefunction` remains True.

`removed()` has no such branch.

## Why it matters

Three `async def` methods are decorated with `@removed`:

- `fmp_data/intelligence/async_client.py:462` — `get_historical_social_sentiment`
- `fmp_data/intelligence/async_client.py:472` — `get_trending_social_sentiment`
- `fmp_data/intelligence/async_client.py:482` — `get_social_sentiment_changes`

After decoration these are no longer coroutine functions. Consequences:

1. **`inspect.iscoroutinefunction()` reports False**, so any introspection that partitions client methods into sync/async silently misfiles all three. `tests/unit/test_deprecated_methods_warn.py` hits this today — it classifies them as sync, so the async sweep never sees them. The test still passes (they raise, which is their contract), but the async guarantee is not actually being asserted for them.
2. **`RemovedEndpointError` raises at call time rather than at `await`.** `asyncio.create_task(client.get_trending_social_sentiment(...))` and `asyncio.gather(...)` raise while *building* the argument, not as a failed task — different from every other async method on the client.
3. **Type checkers** see a declared `async def` whose runtime object is not awaitable.

The error does still surface either way, so this is a consistency and introspection defect rather than a correctness-of-result one. That is why it is filed rather than hot-fixed.

## Fix

Mirror the branch `deprecated()` already has:

```python
def decorator(func: F) -> F:
    if inspect.iscoroutinefunction(func):

        @functools.wraps(func)
        async def async_wrapped(*_args: Any, **_kwargs: Any) -> Any:
            raise RemovedEndpointError(func.__name__, reason)

        async_wrapped.__fmp_deprecated__ = True
        return async_wrapped

    @functools.wraps(func)
    def wrapped(*_args: Any, **_kwargs: Any) -> Any:
        raise RemovedEndpointError(func.__name__, reason)

    wrapped.__fmp_deprecated__ = True
    return wrapped
```

Then tighten `test_deprecated_methods_warn.py` so the async sweep asserts a non-zero count of `@removed` async methods, otherwise the regression reappears unnoticed.

## Provenance

Pre-existing; **not** introduced by #169. The `fmp_data/helpers.py` diff in that PR only adds the three `__fmp_deprecated__` stamps. Found while reviewing #169.

MehdiZare:
Closes #170

`removed()` returned a plain synchronous wrapper regardless of what it decorated, while its sibling `deprecated()` in the same file branches on `inspect.iscoroutinefunction(func)` and returns an `async def` wrapper — its docstring says so explicitly. `removed()` never got the same treatment.

## The consequence is real, not theoretical

A sync wrapper raises when the method is **called**; an async one raises when it is **awaited**. That difference is invisible in isolation and destructive in composition:

```python
await asyncio.gather(removed_endpoint(), other())
```

Before the fix, evaluating `removed_endpoint()` raised **while building `gather`'s argument list** — so `gather()` was never called and **`other()`'s body never ran at all**. A caller batching a removed endpoint alongside live ones silently lost the live work too.

After:

```
iscoroutinefunction(dead): True
  sibling body RAN
raised: RemovedEndpointError
```

The sibling completes and the exception surfaces at `await` time, where it belongs. Called out in the CHANGELOG, since anyone relying on the early raise sees a timing change.

## A correction to the issue's premise

#170 assumed the three async `@removed` methods were being misfiled by `tests/unit/test_deprecated_methods_warn.py`'s sync/async split. They were not: that test used `inspect.iscoroutinefunction(inspect.unwrap(method))`, and `functools.wraps` sets `__wrapped__` to the original async function — so `unwrap` saw straight through the broken wrapper and classified them correctly. They were passing **incidentally**, by way of a workaround nobody knew was load-bearing.

The `unwrap` is now removed, since the wrapper is honest directly, and a `removed_async_checked > 0` floor was added so a future regression cannot hide in that bucket again.

## Verification

```
make lint          clean
unit (extras)      1253 passed,  4 skipped   (was 1247)
unit (core-only)   1063 passed, 11 skipped   (was 1057)
```

Mutation-tested by disabling the new branch (`if False:`), which failed **4** tests: the two new `test_helpers.py` guards, the gather-sibling test, and — unplanned — the pre-existing `checked > 25` floor in the async sweep, which dropped to 23. That last one is the floor from #169 doing exactly its job.

The two decorators are kept structurally parallel rather than factored into a shared helper; at ten lines the duplication reads more clearly than the indirection would.


MehdiZare:
## Code review response

Reviewed against #170. Verdict: **ready to merge**, no Critical or Important findings. Three Minor items, all resolved below.

Everything the review checked was verified empirically rather than read: coroutine-ness (`iscoroutinefunction` → `True`), awaitability, the sync path, `functools.wraps` metadata (`__name__`, `__doc__`, `__wrapped__`, `inspect.signature()`, the `__fmp_deprecated__` stamp), mypy, ruff, and the full unit suite (1253 passed, 4 skipped).

**The new guards are not vacuous.** Reverting `helpers.py` to `32e8815` in a scratch copy fails 4 tests and drops the sweep to 23 async methods reached — matching the CHANGELOG's numbers exactly.

**Every decorator with this wrapper shape was checked, not just the one #170 names.** `grep functools.wraps fmp_data/` yields two sites: `helpers.py` (both branches fixed) and `logger.py:422` (`log_api_call`), which is applied only at `base.py:337` on the **sync** `request()` — `request_async()` is undecorated, so it is not a live bug.

### Findings addressed

**Minor 1 — assert ordering (fixed in 14f233c).** The `removed_async_checked > 0` floor names the exact #170 regression, but the coarser `checked > 25` floor tripped first in that scenario and reported a vague *"is the sweep working?"*. The two asserts are now reordered so the specific diagnosis fires. Verified by mutation: against unfixed `helpers.py` the sweep now fails with the #170 message rather than the generic one.

**Minor 2 — `@deprecated` async `stacklevel` attribution → filed as #177.** `helpers.py:40` warns from inside the coroutine body, so `stacklevel=2` resolves to the frame that *resumed* the coroutine and the warning is blamed on `asyncio/events.py` instead of the user's call site. Measured: sync path reports the caller's file, async path reports `events.py`. This is **pre-existing and untouched by this diff**, and the obvious fix reintroduces exactly the `iscoroutinefunction` lie #170 is about — so it belongs in its own change, not here.

**Minor 3 — async generator functions get the sync wrapper** (`iscoroutinefunction` is `False` for `async def … yield`). There are **zero** async generators in `fmp_data/` today, so this is theoretical; folded into #177 as a related note rather than given its own issue.

### Deliberately not changed

Keeping the two decorators structurally parallel instead of factoring a shared helper. The duplication makes it visible that `deprecated` calls through and `removed` does not — a shared `_wrap_preserving_coroutineness` would hide that distinction.

Raise-on-await (not raise-on-call) is the correct semantics and is explicitly asserted: a removed endpoint fails like every other async client method rather than during expression evaluation.

PR #176, opened 2026-08-09
fix(mcp): make validate, generate and the loader agree (#160, #161, #162, #163, #164)

### LINKED ISSUE #160: fmp-mcp generate rejects bare tool keys that validate and the loader both accept
## What

`fmp-mcp generate --tools <bare key>` reports the key as unknown and drops it. Every other entry point accepts the bare form.

```
$ fmp-mcp generate out.py --no-defaults --tools profile
Warning: Unknown tool 'profile', skipping
Manifest saved to: out.py
Total tools: 0
```

`profile` is a perfectly good manifest entry — `company.profile` is the only spec claiming it, so `register_from_manifest` resolves it and `fmp-mcp validate` blesses it. Only `generate` refuses.

## Why it matters now

Before #149 this was merely inconsistent. After it, it is inconsistent *with a single shared resolver sitting right there*: `resolve_tool_spec(spec, key_to_spec)` answers "what does this entry resolve to" for both forms, and the loader, `validate` and the example-manifest guards all consume it. `generate` is the one place that still does its own membership test:

```python
# fmp_data/mcp/cli.py::_validated_specs
if tool not in available_specs:
    print(f"Warning: Unknown tool '{tool}', skipping", file=sys.stderr)
```

`available_specs` holds only fully qualified specs, so any bare key misses.

The failure is quiet in the worst way: exit code 0, a file written, `Total tools: 0`. A user who passes several bare keys gets an empty manifest and a warning line per key, and nothing says the keys were fine and the *tool* was wrong.

## Suggested fix

Route `_validated_specs` through `resolve_tool_spec` like the other three call sites:

- **resolved** — write the resolved full spec into the manifest (writing the qualified form is the right output regardless, since it is unambiguous under either name style)
- **ambiguous** — say so, naming both candidates, rather than "unknown"; `crypto_quotes` today produces the same misleading message
- **unknown** — keep the current warning, which is then actually true
- **deprecated** — an explicit ask, so honour it, but say it is deprecated and name the replacement (consistent with how `generate` now reports its automatic deprecation drops)

Worth deciding at the same time: should `generate` exit non-zero when an explicit `--tools` selection resolves to nothing? Writing an empty manifest and exiting 0 is arguably the deeper bug here.

## Not in #151

#151 unified resolution across the loader, `validate` and the guards, and deliberately left `generate`'s selection path alone to keep that change reviewable. Filing this so the fourth call site is not forgotten.

## Acceptance

- `fmp-mcp generate out.py --no-defaults --tools profile` writes a manifest containing `company.profile`
- `--tools crypto_quotes` reports an ambiguity naming `alternative.crypto_quotes` and `batch.crypto_quotes`, not "unknown"
- A guard asserts `generate`'s classification of an entry matches `resolve_tool_spec`'s, so this becomes the fourth consumer of the one rule rather than a fourth implementation

### LINKED ISSUE #161: fmp-mcp validate exits 0 for manifests registration refuses
## What

`fmp-mcp validate` prints `Manifest is valid with N tools` and exits 0 for manifests that `register_from_manifest` raises on.

```
$ cat typo.py
TOOLS = ["company.profil"]
$ fmp-mcp validate typo.py
Warning: Unknown tools found: company.profil
Manifest is valid with 1 tools
$ echo $?
0
```

Same for an ambiguous bare key (`crypto_quotes`), and for name clashes. Each of those is a hard `RuntimeError` at server start:

- unknown → `Tool key 'company.profil' not found in available tools`
- ambiguous → `Tool key 'crypto_quotes' is ambiguous...`
- clash → `Duplicate tool keys detected for MCP tool names: crypto_quotes`

The warnings go to stderr; the verdict line and the exit code both say fine.

## Why it matters

The exit code is the part a machine reads. In CI, `fmp-mcp validate manifest.py` is a green check on a manifest that cannot start a server — precisely the failure mode #149's own description warns about:

> a `validate` that blesses manifests the loader then refuses turns a startup error into a trusted green check

#149 removed the *drift* between validate's rule and the loader's; both now consume `resolve_tool_spec` and agree on what each entry resolves to. #151 additionally taught validate to report name clashes. What remains is that validate reports its findings and then reports success anyway.

## Why it is not fixed in #151

Exiting non-zero is a CLI contract change. Anyone with `fmp-mcp validate` in a pipeline today gets a green check; after the fix, a manifest with a typo'd or ambiguous entry fails their build. That is the right outcome, but it is a breaking change to the command's behaviour and wants to be announced as one, not slipped in as a side effect of a resolver refactor.

## Proposal

`validate_manifest` returns `False` when `unknown`, `ambiguous`, `duplicates` or `collisions` is non-empty — i.e. exactly the conditions under which registration raises.

**Deprecated stays a report, not a failure.** It resolves today, and "is my manifest future-proof?" is a question someone should be able to ask without failing their build. This distinction is already written into `_report_manifest_findings`.

Worth settling at the same time:

- Should there be a `--strict` flag that also fails on deprecated, for people who want to gate on it?
- Should the summary line change? `Manifest is valid with 1 tools` printed alongside a fatal warning is contradictory regardless of the exit code — and note the `1 tools` pluralisation bug while touching it.
- `fmp-mcp generate` has the mirror-image problem: an explicit `--tools` selection that resolves to nothing writes an empty manifest and exits 0 (see #160).

## Acceptance

- `fmp-mcp validate` exits non-zero for unknown, ambiguous, duplicate and clashing entries
- exits 0 for a deprecated-but-resolvable manifest, still printing the migration table
- a guard asserts the exit code agrees with whether `register_from_manifest` raises on the same manifest, so the two cannot drift apart again
- CHANGELOG records it as a breaking change to the CLI contract

### LINKED ISSUE #162: Duplicate tool names go undetected under FMP_MCP_TOOL_NAME_STYLE=spec
## What

`_validate_tool_names` returns immediately for any style other than `key`:

```python
# fmp_data/mcp/tool_loader.py
def _validate_tool_names(resolved_specs, name_style):
    if name_style != "key":
        return
```

So the duplicate check exists only under the default style. Under `spec`, a manifest naming one tool twice registers it twice with no error:

```
TOOLS = ["profile", "company.profile"]     # both resolve to company.profile
FMP_MCP_TOOL_NAME_STYLE=spec               # registers as ['company.profile', 'company.profile']
```

Whether the MCP server object then rejects the repeated name or silently keeps the last registration is server-dependent. Either outcome is bad, and neither is announced.

## Why the early return exists, and why it is too broad

The `key` style is where two *different* tools can want one name — `alternative.crypto_quotes` and `batch.crypto_quotes` both advertise as `crypto_quotes`. Switching to `spec` is the documented fix for exactly that, so skipping the check there looks right.

But it conflates two things:

- **collision** — two different specs, one advertised name. Genuinely impossible under `spec`, correct to skip.
- **duplicate** — one spec listed twice, reachable because a manifest accepts both the bare key and the qualified form for the same tool. Unaffected by the name style: `company.profile` twice is `company.profile` twice however names are derived.

The early return skips both, so the style that exists to *fix* name problems is the one with no name checking.

## Suggested fix

Compute advertised names under the effective style, then fail on any repeat:

```python
names = [sem_key if name_style == "key" else full_spec for full_spec, _, sem_key in resolved_specs]
duplicates = sorted({n for n in names if names.count(n) > 1})
```

Under `key` this reproduces today's behaviour exactly. Under `spec` it newly catches the duplicate case while still permitting the collision case. The error message needs to branch: suggesting `FMP_MCP_TOOL_NAME_STYLE=spec` is useless advice to someone already using it.

## Context

Found while reviewing #151, which taught `fmp-mcp validate` to report both cases. `validate` computes names under the effective style and therefore already distinguishes them correctly — this issue is the loader catching up with the validator. Note the direction: #149 was about `validate` blessing what the loader refuses; this is the loader accepting what `validate` warns about. Same class of divergence, opposite sign.

## Acceptance

- Registering `["profile", "company.profile"]` raises under **both** name styles
- Registering both `crypto_quotes` specs still succeeds under `spec` and still raises under `key`
- The error names the offending tool and gives advice appropriate to the style in effect
- A guard asserts the loader's verdict matches `fmp-mcp validate`'s clash report for the same manifest under both styles

### LINKED ISSUE #163: fmp-mcp list output is not copy-pasteable into a manifest
## What

`fmp-mcp list` is the documented way to discover what to put in a manifest. Two of its four formats do not show a usable manifest entry.

**`table` (the default) truncates the Tool Spec column:**

```
alternative.commo…
company.historica…
company.historica…
```

Those last two are `company.historical_price` (deprecated) and `company.historical_prices` (its replacement). Rendered, they are the same string. The `[DEPRECATED -> ...]` marker is part of the truncated cell, so it is cut off too — the one row a user most needs to distinguish is the one the format flattens.

**`tree` prints method names, not tool keys:**

```
get_historical_prices [DEPRECATED -> company.historical_prices]
get_historical_prices
```

The actual manifest entry, `company.historical_price` (singular), never appears. A user reading this cannot tell what to write or what to remove, and the deprecated row is labelled with the *replacement's* method name, so it reads as self-referential.

Only `--format list` and `--format json` are complete and copy-pasteable.

## Why it matters

`docs/mcp/configurations.md` sends users to `fmp-mcp list` to discover tools, and #136's deprecation cycle relies on users being able to see which key is deprecated and what replaces it. Both defaults defeat that: the format shown when you type the bare command is the one that hides the answer.

## Suggested fix

- **table**: do not truncate the spec column — it is the primary key of the output and typically under 40 characters. Truncate the description instead (already the case) and widen or wrap the spec. Consider a separate `Deprecated` column so the marker cannot be cut.
- **tree**: label leaves with the tool key, since the branch already carries the client — `historical_price` under `company` reads as `company.historical_price`. Keep the method name as trailing detail if useful.

## Context

Found while verifying #151. Pre-existing on `dev`, untouched by that PR, which is why it is filed rather than folded in.

## Acceptance

- No format renders two distinct tools identically
- The deprecation marker and its replacement survive in every format
- `company.historical_price` and `company.historical_prices` are visibly distinct in `table` and `tree`
- A test asserts every format emits, for each tool, a string containing the full spec

### LINKED ISSUE #164: Two deprecation mechanisms, and fmp-mcp generate only honours one
## What

There are two independent ways an endpoint is "deprecated" in this codebase, and `fmp-mcp generate` filters on one of them.

1. **`fmp_data.mcp.tools_manifest.DEPRECATED_TOOLS`** (#136) — a duplicate tool *key* for a method that still works. Three entries. `generate` omits these and names each with its replacement.
2. **`EndpointSemantics.deprecated`** (#137) — the endpoint itself returns `[]` without calling upstream. Three entries: `intelligence.earnings_confirmed`, `intelligence.earnings_surprises`, `intelligence.stock_news_sentiments`. `generate` includes all three.

The two are deliberately disjoint and a test asserts they stay so — see the #137 CHANGELOG entry, which is explicit that the mechanisms must not be merged. That reasoning is sound for *resolution*: one is a rename, the other is a dead endpoint, and the fix differs.

It does not obviously carry over to *generation*. A generated manifest today advertises three tools that return an empty list on every call, and `stock_news_sentiments` describes itself as "DEPRECATED and non-functional ... always returns an empty list". An LLM handed that tool gets an empty success, which is #137's original complaint — the same defect that motivated hiding them from the LangChain vector store.

Secondary point: the generated header reads `Omitted because they are deprecated`, which a reader may take as covering everything deprecated. It covers what was omitted, which is accurate but easy to misread when six things are deprecated and three are omitted.

## The actual question

Should `generate` filter `EndpointSemantics.deprecated` too?

**For:** the LangChain side already hides them for exactly this reason; a manifest full of tools that cannot return data is not a good default; `generate`'s stated goal is a manifest that works.

**Against:** MCP addresses tools as `<client>.<key>` and #137 deliberately kept the catalog at 223 with all three resolvable, so omitting them from `generate` is a step toward a removal that has not been agreed. Unlike the `DEPRECATED_TOOLS` case there is no replacement to name, so the header note would be "omitted, nothing replaces it" — closer to a removal announcement than a migration hint.

My inclination is to omit them with a distinct header section ("omitted because they return no data") rather than reuse the deprecation wording, and to leave resolution untouched so an explicit manifest entry still works. But this changes what a default generated manifest contains, so it wants a decision rather than a patch.

## Not a regression

Both mechanisms behave as their PRs specified. This is the seam between them, visible only now that `generate` has become deprecation-aware at all.

## Acceptance (if we go ahead)

- A decision recorded in the CHANGELOG either way
- If omitted: its own header section with its own wording, resolution unchanged, and a guard that the two mechanisms remain distinguishable in `generate`'s output
- If not omitted: a note in `docs/mcp/configurations.md` that a generated manifest may include endpoints that return no data, and why

MehdiZare:
Closes #160
Closes #161
Closes #162
Closes #163
Closes #164

Five issues, one root cause: **`fmp-mcp`'s entry points each had their own opinion about what a valid manifest is.**

The worst of them, #161: `validate` printed "Manifest is valid with N tools" and exited 0 for manifests `register_from_manifest` raises on. A validator that disagrees with the thing it validates is worse than none, because it converts a startup error into a trusted green check.

## Three reimplementations of the resolution rule, not one

#151 introduced a pure `resolve_tool_spec()` precisely so `validate` and the loader would stop reimplementing it. Two more copies survived: `_manifest_name_clashes` and `_name_collisions`, each with its own duplicate-vs-collision split. That split now lives once, in `tool_loader.advertised_names` + `split_name_clashes`, consumed by the loader, `validate` and `generate`.

## The guard that makes #161 structurally impossible

A table of **11 manifests × 2 name styles** asserting `validate`'s verdict is the exact inverse of whether `register_from_manifest` raises — with a floor requiring ≥4 raising and ≥4 non-raising cases so it cannot go vacuous. Fixing #161 once was never the point; a fourth reimplementation now fails the build.

## #164 — one table that was three entries short

`WITHDRAWN_TOOLS` (19) was already a strict subset of `EndpointSemantics.deprecated` (22), with `DEPRECATED_TOOLS` (3 aliases) disjoint from both. The gap was `intelligence.earnings_confirmed`, `earnings_surprises`, `stock_news_sentiments` — and all three were probed against the live API and **404**. So they are genuinely withdrawn; filing them there completes the table rather than merging two concepts.

```
WITHDRAWN 19 -> 22    DEPRECATED 3 (disjoint, unchanged)    DEFAULT_TOOLS 137 (unchanged)
generated catalog manifest 201 -> 198
```

None was ever a default, so **a default server is unaffected and this is not breaking for server users.**

The invariant guard derives the flagged set by walking the mapping tables rather than hard-coding it, asserts set equality, and carries a `>= 20` floor so a broken walk cannot satisfy it against an empty set. A third test pins the three by name, so the equality cannot be met by deleting the *flag* instead of adding the entry.

## Breaking CLI changes — three, and please note the third

1. **`validate` exit code** — now non-zero for manifests registration refuses.
2. **`generate` exit code** — now non-zero when an explicit `--tools` selection resolves to nothing, instead of writing an empty manifest and reporting success.
3. **`list` output shape** (#163) — new `Deprecated` column; tree leaves are now specs. **Anything parsing `fmp-mcp list --format table` will see different columns.** Neither of us called this breaking up front; it is, and it is in the CHANGELOG and `docs/mcp/configurations.md` as such.

On (2), `generate` now **writes nothing** on total failure rather than clobbering a good manifest with an empty one — verified:

```
Error: no tools selected, so <path> was not written.
None of the 1 requested tool resolved:
  nonsense.nope: unknown -- names nothing in the catalog
existing file preserved: True
```

Also: `generate_manifest()` returns `bool` (was `None`), and `_validated_specs` gained a `failures` out-param.

## Verification

```
make lint          clean
unit (extras)      1322 passed, 10 skipped
unit (core-only)   1132 passed, 17 skipped
```

Every guard is mutation-verified — each defect was reintroduced and the failure confirmed: the `spec`-style early return → 3 failures, old tree leaves → 3, pre-#164 `tools_manifest.py` → exactly the 3 retirement guards, pre-fix table columns → the ellipsis assertion.


MehdiZare:
## Code review response

Reviewed against #160–#164, with all three entry points exercised over 10 manifest shapes × 2 name styles, every guard mutation-tested, and the suite run in a **core-only** environment. Verdict: **merge with fixes** — no Critical. Both Important findings were real, both are fixed in `f48a049`.

The review confirmed the PR's central claims rather than taking them: `resolve_tool_spec` has exactly four call sites with no surviving parallel implementation; `validate`'s verdict is the exact inverse of whether registration raises in **every** cell tested, including the subtle one where a collision is fatal under `key` and fine under `spec`; and the `#164` phantom-count guard **is not safe-by-accident** — rewriting `_semantics_flagged_specs` as the naive module walk its own docstring warns about (including the `lc.mapping` re-export trap) fails two tests.

### Important 1 — `generate` exited 0 writing a manifest `validate` exits 1 on

The collision shape, and a live counterexample to the PR's headline claim:

```
$ fmp-mcp generate coll.py --no-defaults --tools alternative.crypto_quotes batch.crypto_quotes  → EXIT 0
$ fmp-mcp validate coll.py                                                                       → EXIT 1
```

Fixed, but **not** by making collisions unconditionally fatal — that would have been the mirror-image bug. `_name_collisions` is deliberately fixed to `key` style, while `validate` judges under the *active* style, so an unconditional refusal would make `generate` exit 1 under `spec` where `validate` exits 0. The refusal is now gated on the active style:

| | `key` (default) | `spec` |
|---|---|---|
| `generate` | **exit 1**, nothing written | exit 0, both sides written |
| `validate` | exit 1 | exit 0 |

Neither side is ever thinned — the selection is refused *whole*, with the escape hatch named, so the user makes the call rather than `generate` picking a winner. Verified end to end; `generate && validate` now agrees in all four cells.

### Important 2 — the announced breaking change only held under `--no-defaults`

```
$ fmp-mcp generate out.py --tools profil
Warning: Unknown tool 'profil', skipping.
Manifest saved to: out.py / Total tools: 137     → EXIT 0
```

`_add_defaults` ran *before* the emptiness check, so a selection in which nothing the user asked for resolved was rescued by 137 defaults they never asked for, `failures` was discarded, and the command reported success. And you're right that no guard covered it: every case in `TestGenerateRefusesToWriteAUselessManifest` passed `include_defaults=False`, so the documented sentence was pinned only on the opt-in path.

Taken the stronger option — an explicit selection's emptiness is now judged **before** defaults are considered, so this is fatal regardless. The defaults are not an answer to an ask that named only tools which do not exist. Guarded with `include_defaults=True`.

### Minor items

- **3 — `list` output change missing from the docs.** Correct, and the PR body claimed otherwise. Now documented in `docs/mcp/configurations.md` §3, including that `--format json` is the stable interface for scripts.
- **4 — the width guard ran at `COLUMNS=400`.** Good catch: at that width rich never has to shrink, so the guard cannot distinguish a lossless fold from a lossy cut. Rather than claim copy-pasteability that isn't true at 80 columns, the fold is now documented as a known rough edge **and** pinned: a second guard reassembles the spec column across folded lines at `COLUMNS=80` and requires both `#163` specs back intact. Verified it catches a hard truncation using no ellipsis character, which the existing ellipsis check misses.
- **5 — `_get_tool_name_style` reached across module boundaries.** Promoted to `tool_name_style()`; the private name is kept as an alias.
- **6 — `TOOLS = []` asymmetry.** Added to the validate table with the reasoning: `validate` answers "can this start a server" (an empty manifest can), while for `generate` the emptiness *is* the failed ask.
- **`--strict` (#161's open sub-question)** — now recorded as declined in the CHANGELOG rather than left hanging. The fix above makes the exit code itself the verdict for every shape that cannot start a server; a flag to opt into correctness implies the default is something else.

### One consequence worth flagging

`test_explicit_tools_keep_both_sides_and_get_guidance` asserted the old contract, so it was **rewritten, not deleted** — it now pins the write-both-sides behaviour under `spec`, and a new sibling pins the refusal under `key`. Both halves of "never thinned" remain covered.

### Mutation evidence for the new guards

| mutation | caught by |
|---|---|
| drop the early emptiness check (defaults rescue returns) | `test_defaults_do_not_rescue_an_unresolvable_explicit_selection` |
| collisions always fatal (ignore active style) | `test_the_same_collision_is_written_under_spec_style` |
| collisions never fatal (the original bug) | `test_a_collision_under_key_style_is_not_written` |
| hard-truncate the spec column, no ellipsis | the reassembly assertion at `COLUMNS=80` |

Full suite: **1326 passed, 10 skipped**; ruff and mypy clean. The one remaining `mypy` error (`fmp_data/mcp/_compat.py:31`) is pre-existing and byte-identical at base.

Issue #180, opened 2026-08-09
test_the_spec_column_is_never_ellipsised_on_a_default_terminal silently requires rich


MehdiZare:
Pre-existing on `dev` (from #176/#163). Found while verifying an unrelated branch in a genuinely extras-free virtualenv.

## What

`tests/unit/test_mcp_entry_point_coherence.py::TestListOutputIsCopyPasteable::test_the_spec_column_is_never_ellipsised_on_a_default_terminal` fails in an environment without `rich`:

```
AssertionError: company.historical_price is not recoverable from the 80-column table;
the spec column is losing characters, not just wrapping
assert 'company.historical_price' in ''
```

Reproduced on `origin/dev` directly, so this is not introduced by any open branch.

## Why

The test asserts on `rich`'s box-drawing output — it splits rendered lines on `│` to reassemble the spec column. Without `rich`, `_print_tools_table_format` falls back to `_print_simple_table`, which emits plain text containing no `│`. The reassembled column is therefore the empty string and the membership assertion fails.

The module has no `importorskip` for `rich`, and the test does not branch on `HAS_RICH`.

## Why CI is green

CI installs `rich`, so the test exercises the path it was written for and passes. The failure only appears in a真 core-only install — which is the configuration the project has been careful about elsewhere, and which has caused four separate incidents in this issue series.

## Scope

Test-only. No runtime behaviour is affected: `fmp-mcp list` degrades correctly to the simple table without `rich`, and that fallback is deliberate (`HAS_RICH` at `fmp_data/mcp/cli.py:16-23`).

## Options

1. `pytest.importorskip("rich")` on the test, since it is specifically about rich's ellipsis behaviour and has no meaning without it.
2. Branch on `HAS_RICH` and assert the equivalent lossless property for the simple-table path — stronger, since the simple table is what an extras-free user actually sees, and nothing currently tests that it does not truncate.

Option 2 is the more useful one: the simple-table renderer is the unguarded path today, and #163's claim ("`list` output is copy-pasteable") should hold for both renderers, not just the pretty one.

## Related

Same class as the `lc.mapping` note recorded in that file's docstring — a test depending on an optional package without saying so. That one was safe by accident; this one is not.

MehdiZare:
Fixed by #182 (merged). The test is now skipped when `rich` is absent.

[...]

PR #181, opened 2026-08-09
feat(lc): derive tool examples and Literal types from valid_values; fold technical date hints (#156, #179)

### LINKED ISSUE #156: Derive LangChain tool parameter examples from valid_values, and surface valid_values in the generated schema
Split out of the review of #150 (see the `PERIOD_HINT` three-way split there).

## The problem

Two sources describe the same parameter and nothing keeps them in sync:

- `ParameterHint.examples` is rendered verbatim into the tool description by `ToolFactory.generate_description` — this is what the LLM reads and imitates.
- `EndpointParam.valid_values` is enforced by `EndpointParam.validate_value` — this is what the client accepts.

`ToolFactory.create_parameter_fields` does **not** narrow the generated field type by `valid_values`, so the enum never reaches the model at all. When the two disagree, the tool advertises a value that fails client-side with a `ValidationError` the model cannot anticipate.

#150 found nine such endpoints, all on `period`, and fixed them by splitting the shared hint three ways (`PERIOD_HINT` / `PERIOD_WITH_FISCAL_HINT` / `FISCAL_PERIOD_HINT`) — one hint per `valid_values` set. That works, but it is the symptom being managed: every future `valid_values` set under a shared parameter name needs another hint constant.

## The fix

1. When a parameter declares `valid_values`, derive its advertised examples from that list rather than from the hint, or intersect the two. The hint keeps owning natural names, context clues and extraction patterns.
2. Emit `valid_values` into the generated pydantic field as a `Literal` (or an enum), so the constraint reaches the model instead of being discovered on rejection.

Either change alone removes a class of defect; together they make the disagreement unrepresentable, and the three `period` hints could collapse back to one.

## Guard already in place

`tests/unit/test_hint_consolidation.py::test_hint_examples_are_within_valid_values` (added in #150) walks all 215 registered endpoints and fails on any example outside `valid_values`. It should keep passing through this refactor, and would become trivially true once examples are derived.

## Note

(2) changes the tool schema every LLM sees, so it is a user-visible change worth a CHANGELOG entry in the shape of #128's.

### LINKED ISSUE #179: Fold technical/mapping.py's local date hints onto the shared DATE_HINTS (#157 option 1)
Fold `technical/mapping.py`'s local `FROM_DATE_HINT`/`TO_DATE_HINT` onto the shared `fmp_data.lc.hints.DATE_HINTS`.

This is #157's "Options" item 1, deferred out of #173 and tracked here so it isn't carried only by a code comment. #173 landed the *guard*; this is the remaining *fix*.

### Current state

`fmp_data/technical/mapping.py` defines its own date hints:

```python
FROM_DATE_HINT = ParameterHint(
    natural_names=["from date", "start date", "beginning"],
    extraction_patterns=[r"from\s+(\d{4}-\d{2}-\d{2})", r"starting\s+(\d{4}-\d{2}-\d{2})"],
    ...
)
```

These describe the same concept as `DATE_HINTS["start_date"]`/`DATE_HINTS["end_date"]` under different names and with different contents. The `#173` detector flags all 36 resulting pairs; they are allowlisted in `KNOWN_SAME_CONCEPT_DIVERGENCES` in `tests/unit/test_hint_consolidation.py` with this issue as the reason.

### Why it was deferred

Folding them changes `natural_names`, `extraction_patterns`, `examples` and `context_clues` for the nine technical-indicator tools. Those fields are rendered into tool descriptions and into the text that gets embedded for semantic search, so this is a **behaviour change to retrieval quality**, not a refactor. #157 scoped it as needing its own before/after review of the nine affected tools rather than riding along with a guard change.

### What the fix involves

1. Replace the two local definitions with `DATE_HINTS["start_date"]` / `DATE_HINTS["end_date"]` (add `DATE_HINTS` to the existing `fmp_data.lc.hints` import).
2. Delete all four entries from `KNOWN_SAME_CONCEPT_DIVERGENCES`.
3. Review the before/after tool descriptions and embedding text for the nine indicator tools.

Verified in a scratch copy that steps 1–2 leave `tests/unit/test_hint_consolidation.py` green (9/9) — the guard's floor counts *comparisons*, not *findings*, precisely so that fixing this does not break it.

### Also worth knowing

`_hint_bindings` sees only module-level bindings plus one level of dict, so the ~20 `ParameterHint(...)` literals written inline inside `parameter_hints={...}` arguments (in `alternative`, `economics`, `fundamental`, `intelligence` mappings) are invisible to both this guard and #135's. Not in scope here, but if those are ever folded too, the guard needs extending to see them.

MehdiZare:
Closes #156
Closes #179

## #156 — the endpoint already knows what it accepts

An enum-constrained parameter reached the LLM with hand-written examples that could drift from what the endpoint actually accepts, and the constraint itself was invisible in the generated schema. A live API audit this week confirmed `valid_values` is authoritative — several endpoints 400 on values outside it.

Examples are now derived from `param.valid_values` when declared (Enum members unwrapped to `.value`), falling back to `hint.examples` otherwise. The pydantic field narrows to `Literal[...]`, so the constraint reaches the schema rather than living only in prose:

```python
period annotation: Optional[Literal['annual', 'quarter', 'FY', 'Q1', 'Q2', 'Q3', 'Q4']]
endpoint valid_values: ['annual', 'quarter', 'FY', 'Q1', 'Q2', 'Q3', 'Q4']
```

## #179 — the last same-concept divergence

`technical/mapping.py`'s local `FROM_DATE_HINT`/`TO_DATE_HINT` now come from the shared `DATE_HINTS`. This was the one real divergence #157's guard found, baselined as out of scope at the time; **`KNOWN_SAME_CONCEPT_DIVERGENCES` is now empty**, which is the point — that guard was built so a fixed entry must be deleted rather than left to rot.

The 9 indicator tools' embeddings move, as intended: `natural_names` gains "since"/"through"/"to"; `extraction_patterns` gains a bare ISO-date pattern; examples shift to 2023/2024 dates; `context_clues` gains "after"/"ending" and loses "start"/"end". Spelled out in the CHANGELOG since natural-language extraction shifts for those tools.

## The guard that needed a second test

The catalog-wide guard (`test_constrained_params_advertise_only_valid_values`, floor 20/29) asserts every constrained param's examples are a subset of its `valid_values`. But **no real hint drifts today**, so reverting the fix did not make it fail — it was correct and vacuous at the same time.

A second, deterministic test manufactures a drifted hint directly. Verified: reverting either the derivation or the `Literal` narrowing fails *that* test, not the catalog sweep. Both are kept — the sweep catches future real drift, the deterministic one proves the mechanism works now.

That distinction has bitten this issue series repeatedly, so it is worth naming: a guard that cannot fail today is not yet a guard.

## Verification

```
make lint            clean
mypy (nox)           clean, 204 files
unit (extras)        1346 passed, 10 skipped
unit (core-only)     1152 passed, 17 skipped
```

Both fixes mutation-tested: re-adding the baseline entries with the fold applied fails the stale-entry assertion; reverting the fold with the baseline empty fails with 36 unallowlisted candidates.

## Note on the core-only count

One pre-existing failure appears in a genuinely extras-free venv, on `dev` as well as here: `test_the_spec_column_is_never_ellipsised_on_a_default_terminal` silently requires `rich`. Filed as **#180**; not introduced by this branch and reproduced on `origin/dev` directly.


MehdiZare:
## Review follow-up

Independent code review (no prior inline threads) — Important findings addressed in `c859e90`:

1. **Manufactured #156 guard now goes through production wiring** — `create_parameter_fields` → `create_model`, not only `get_examples_for_param` / `get_field_type`. Asserts examples, description, `ValidationError` on drifted values.
2. **#150 vs #156 pipeline clarity** — CHANGELOG, hint/consolidation docs, and comments now distinguish ToolFactory schema derivation (`valid_values`) from the embedding/hand-written hint content guard (#150).

**Deferred (isolated / broader):**
- Optional: share one examples helper for registry embeddings + tool schema (future issue if we want embeddings and tools to share derivation).
- `claude-review` CI red is infra (missing OAuth secret) — tracked in #184, not a product defect.

Tests: `test_tool_schema` + `test_hint_consolidation` green locally.

PR #182, opened 2026-08-09
fix: attribute async deprecation warnings to the caller; surface retirement in docs and list (#177, #178, #158, #180)

### LINKED ISSUE #158: fmp-mcp list does not mark endpoints deprecated via EndpointSemantics.deprecated
Split out of the review of #150 (#137).

`fmp_data/mcp/cli.py:43` annotates deprecation solely from `DEPRECATED_TOOLS`, the table of duplicate *names* for a live method (#136). It knows nothing about `EndpointSemantics.deprecated`, the flag #150 added for endpoints that return `[]` without calling upstream.

So `fmp-mcp list` prints `stock_news_sentiments`, `earnings_confirmed` and `earnings_surprises` with `deprecated: None`, indistinguishable from live tools. Since custom manifests are built by copying keys out of that listing, a user will add a tool that can only ever return an empty result — the same "empty success" #137 removed from the LangChain side.

The keys deliberately stay resolvable (an explicit manifest must keep working without a deprecation window), so this is about labelling, not removal.

## Fix

Have the CLI consult `EndpointSemantics.deprecated` alongside `DEPRECATED_TOOLS`, and render it in all four output formats (`list`, `tree`, `table`, and the `deprecated` field in `--format json`). The two mechanisms mean different things — "renamed, still works" vs "returns nothing" — so the label should distinguish them rather than collapse both into one flag.

Note this changes the shape of `fmp-mcp list --format json` output, which some users script against.

Related guard: `tests/unit/test_mcp.py::TestDeprecationMechanismsStaySeparate` asserts the two sets stay disjoint.

### LINKED ISSUE #177: @deprecated async warnings are attributed to asyncio internals, not the caller
`@deprecated`'s async branch attributes its DeprecationWarning to `asyncio/events.py`, not the caller.

Found while reviewing #175 (which fixes the sibling `@removed` coroutine bug, #170). This one is pre-existing and was deliberately left out of that PR to keep it a clean single-purpose fix.

### Symptom

`fmp_data/helpers.py:40` — the async wrapper calls `warnings.warn(..., stacklevel=2)` from inside the coroutine body. `stacklevel=2` there resolves to the frame that *resumed* the coroutine (`Task.__step`), not the frame that called it. Measured:

```
sync path:  DeprecationWarning at probe.py 85      <- caller's file, correct
async path: DeprecationWarning at events.py 88     <- asyncio internals, wrong
```

### Why it matters

A user calling a deprecated async client method sees the warning blamed on CPython's event loop. They cannot locate their own call site, which is the entire point of a DeprecationWarning. Filters keyed on `module=` in the user's own code also fail to match.

`@removed` is unaffected — it raises, and the traceback carries the real frame.

### Why it isn't a one-liner

The obvious fix (warn eagerly in a sync wrapper that returns a coroutine) reintroduces exactly the `inspect.iscoroutinefunction` lie that #170 was about, so it must not be done that way. Realistic options:

1. Walk the stack past `asyncio` frames to compute a correct `stacklevel`.
2. Use `warnings.warn_explicit` with the caller's filename/lineno captured at decoration time.
3. Accept the limitation and document it.

### Related, non-blocking

Async *generator* functions (`async def ... yield`) get the sync wrapper from both `@deprecated` and `@removed`, because `inspect.iscoroutinefunction` is `False` for them — so `isasyncgenfunction()` on the result is `False` and the wrapper raises/warns at call time rather than on iteration. There are currently **zero** async generators in `fmp_data/`, so this is theoretical. Worth handling in the same pass if option 1 or 2 is taken.

### LINKED ISSUE #178: Concrete argument models carry no .. deprecated:: marker, only the 8 roots do
The ~100 concrete argument models carry no `.. deprecated::` marker — only the 8 roots do.

Found while reviewing #174, which deprecates `Endpoint.arg_model` and the hand-written argument models (#153, #167). Deliberately left out of that PR: the fix is either 100+ mechanical docstring edits or a small metaprogramming change, and neither belongs in a PR whose point is the deprecation mechanism.

### Symptom

#174 added `.. deprecated:: 2.7` to the 8 root classes (`BaseArgModel`, `BaseListArgs`, `BaseQuoteArgs`, `BaseSearchArg`, `BaseExchangeArg`, `QuoteArgs`, `MarketCapArgs`, `BaseSymbolArg`).

The 116 concrete classes users actually name in their own code — `ETFHoldingsArgs`, `TechnicalIndicatorArgs`, `Form13FArgs`, … — inherit the deprecated base but carry no marker of their own. Docstrings are not inherited for documentation purposes, so:

- `help(ETFHoldingsArgs)` shows nothing about the deprecation
- IDE hover shows nothing
- Sphinx renders no deprecation admonition

The runtime `DeprecationWarning` does fire for these classes (on validation, and on `model_json_schema()` after #174's review follow-up), so a user who *runs* the code is told. A user who *reads* the docs before writing the code is not — which is the audience most able to avoid the migration entirely.

### Options

1. **`__init_subclass__` on `DeprecatedArgModel`** that appends a `.. deprecated:: 2.7` paragraph to any subclass `__doc__` lacking one. One place, covers every current and future subclass, no churn. Needs care: `__doc__` is writable on classes but the appended text must not break existing Sphinx directives, and classes with `__doc__ = None` need a base string.
2. **Mechanical edit of ~116 docstrings.** Explicit and greppable, but large and easy to drift from reality.
3. **Accept it** and rely on the runtime warning plus the CHANGELOG.

Option 1 is probably right, and it composes with the existing guard in `tests/unit/test_arg_model_deprecation.py` — that walk already enumerates all 116 classes, so a guard asserting each one's `__doc__` mentions the deprecation is a few lines on top of machinery that exists.

### Scope note

Whatever is chosen must keep **import silent**. `test_importing_an_arg_model_stays_silent` pins that; a docstring rewrite at class-creation time is fine, a warning at class-creation time is not.

### LINKED ISSUE #180: test_the_spec_column_is_never_ellipsised_on_a_default_terminal silently requires rich

MehdiZare:
Closes #177
Closes #178
Closes #158
Closes #180

## #177 — the warning blamed CPython's event loop

`@deprecated`'s async branch called `warnings.warn(..., stacklevel=2)` from inside the coroutine body, so `stacklevel=2` resolved to the frame that *resumed* the coroutine (`Task.__step`), not the one that called it:

```
sync path:  DeprecationWarning at probe.py 85    <- correct
async path: DeprecationWarning at events.py 88   <- asyncio internals
```

A user calling a deprecated async client method could not locate their own call site — the entire point of a `DeprecationWarning` — and `filterwarnings` rules keyed on their own module never matched. Verified after:

```
  probe177.py:12    (sync)
  probe177.py:13    (async)
```

Both now name the caller's file. This is the same class as the `stacklevel=3`→`4` fix in #147; the guard asserts the warning's **filename is the caller's**, because asserting merely that a warning fired is exactly what let this survive.

## #178 — only the roots were marked

Concrete argument models carried no `.. deprecated::` marker; only the 8 root classes did. Anyone reading `ETFHoldingsArgs` saw nothing to suggest it is going away in 3.0.

## #158 — the third retirement concept was invisible in `list`

There are three deliberately distinct concepts: `DEPRECATED_TOOLS` (an alias that still works, removed 3.0), `WITHDRAWN_TOOLS` (FMP no longer serves the endpoint), and `EndpointSemantics.deprecated` (the LangChain-side flag, which #176 made equal to `WITHDRAWN_TOOLS`). #176 taught `list` to mark the first. It now labels withdrawn endpoints too, so all three are visible without merging any of them.

## #180 — a test that silently required `rich`

Found while verifying an unrelated branch in a genuinely extras-free venv, and reproduced on `origin/dev` directly. `test_the_spec_column_is_never_ellipsised_on_a_default_terminal` asserts on `rich`'s box-drawing output by splitting on `│`; without `rich` the code correctly falls back to a plain table containing none, so the reassembled column was empty and the assertion failed. CI installs `rich`, so it was green there.

Now skipped when `rich` is absent. Fixed here rather than left open because this branch was already in the file.

## Verification

```
make lint          clean
unit (extras)      1368 passed, 10 skipped
unit (core-only)   1164 passed, 18 skipped
```

Core-only is run separately because CI installs no extras — a gap that has caused four separate incidents in this issue series, and which is what surfaced #180.


MehdiZare:
## Review follow-up

Independent code review — Important + polish addressed in `4a7a775`:

1. **`_first_caller_frame` now skips `uvloop` / `uvloop.*`** (same class as asyncio resume frames under uvicorn/FastAPI), with a unit test that injects a fake frame chain (no real uvloop dependency).
2. **Comment fix** for `filterwarnings(module=...)` (matches module name, not filename).
3. **CHANGELOG** one-liner for #180.

**Deferred:** other event loops (`trio`, etc.) can be added to `_INTERNAL_LOOP_MODULES` if they show up in production stacks. CI `claude-review` failure is #184.

Tests: `tests/unit/test_helpers.py` — 24 passed.

MehdiZare:
## Rebase note

Rebased onto latest `dev` after #181/#183 landed. Conflict was only in `CHANGELOG.md` (Unreleased entries from those PRs vs this PR’s #158/#177/#178/#180 notes) — both sides kept. Force-pushed `b19739b`.

Issue #166, opened 2026-08-09
market.search-name: Endpoint.name is hyphenated, equals its path, and is load-bearing as a cache key


MehdiZare:
Found while adding row-set equality to the `docs/api/endpoints.md` guard in #155/#146.

## What

```python
# fmp_data/market/endpoints.py:111-112
    name="search-name",
    path="search-name",
```

Two oddities in one definition:

1. **It is the only endpoint whose `name` uses a hyphen.** Every other endpoint in the catalog uses `snake_case` for `name` and reserves hyphens for `path`. `docs/api/endpoints.md` documented it as `search_name`, matching the convention rather than the code — the documentation was simply wrong, and the count-based doc guard could not see it because the count was right either way.
2. **Its `name` is identical to its `path`.** Everywhere else the two are independent: `crypto_news_symbol` has path `news/crypto`.

## Why it is not a trivial rename

`Endpoint.name` is not cosmetic. It is the cache-key prefix and the argument to the TTL lookup:

```
fmp_data/base.py:405   cache_key = self._build_cache_key(endpoint.name, query_params)
fmp_data/base.py:406   cached = self._cache_get(endpoint.name, cache_key)
fmp_data/base.py:457   ttl = self._get_cache_ttl(endpoint.name)
```

Renaming it changes the cache namespace for that endpoint, so any persisted cache silently misses on the first request after the change, and any TTL configuration keyed on the old name stops applying. That is a small, real behaviour change, and it is why #155 corrected the *document* to say `search-name` rather than correcting the code to say `search_name`.

## Suggested direction

Rename `name` to `search_name` for consistency, in a change that stands on its own and notes the cache-key implication in the changelog. Check first whether any TTL configuration or cache-warming code keys on the literal string.

The doc is now guarded row for row (name *and* path), so whichever side changes, the other is forced to follow.

MehdiZare:
A second instance of the same convention drift, found in review of #154:

```python
>>> from fmp_data.institutional.endpoints import FAIL_TO_DELIVER
>>> FAIL_TO_DELIVER.name, FAIL_TO_DELIVER.path
('fail_to_deliver', 'fail_to_deliver')
```

Underscores in a `path`, where every other path in the catalog uses hyphens — and, as with `search-name`, `name` and `path` are the same string. Pre-existing on `dev`; neither #154 nor #155 touched it.

[...]

PR #183, opened 2026-08-09
refactor(models): derive required from list membership; warn on unmatched ttl_overrides (#165, #171)

### LINKED ISSUE #165: EndpointParam.required should be derived from list membership, not stored
Split out of #144, which #155 fixes by option 1 (make the flags agree with the list). This is option 3, the one #144 itself calls "the honest one".

## What

`EndpointParam.required` is stored on every parameter — 546 `required=` assignments across the catalog — and read in exactly one place:

```
fmp_data/models.py:221    if self.required:  # inside validate_value
```

That branch is only reachable for a parameter in `mandatory_params`, because `validate_params` short-circuits first:

```python
# Skip None values for optional params
if value is None and param.name not in mandatory_names:
    continue
```

So for an optional parameter the flag can never be consulted, whatever it says. Requiredness is decided entirely by which list the parameter is declared in.

## Why it matters

#155 reconciled the 14 flags that contradicted their list and added `tests/unit/test_param_required_consistency.py` to keep them reconciled. That guard is the right fix for the symptom, but it exists to keep two sources of truth agreeing when there only needs to be one. Every new endpoint definition still has to set a field that nothing reads, and can still set it wrong.

## Suggested direction

Remove the field and derive requiredness from list membership at the one read site. `validate_value` would need the answer passed in, or the check would move up into `validate_params` where `mandatory_names` already exists.

This touches every endpoint definition in the package, which is why it did not belong in #155. It also makes `test_param_required_consistency.py` redundant — the contradiction becomes unrepresentable rather than merely guarded.

### LINKED ISSUE #171: Follow-ups from #169 review: e2e exemption keying, withdrawn-header assertions, CHANGELOG count reconciliation
Deferred from the code review of #169. None is a live defect; each is a latent gap or a clarity problem. Grouped because they are all small and all in the same area.

## 1. `tests/e2e/test_live_signatures.py` — exemption keyed by bare attribute name

`_deprecated_endpoints()` keys the 404-sweep exemption set by the bare endpoint attribute name rather than by `(client, attr)`.

Verified: **zero** cross-client attribute-name collisions exist across all 275 declarations today, so nothing is wrong right now. But if one ever appears, a *live* endpoint that happens to share a name with a dead one is silently exempted from the 404 sweep — the exemption stops being earned, which is the exact property `test_deprecated_endpoints_are_really_dead` was written to guarantee.

`name_by_endpoint` already carries the client, so keying on the pair costs nothing.

## 2. No withdrawn counterpart to `test_deprecated_drops_name_their_replacement`

`tests/unit/test_mcp_cli.py` asserts that each *deprecated* header line names its replacement. There is no equivalent for withdrawn: nothing asserts that each withdrawn line names its successor, or says "FMP publishes no replacement" for the four that have `None`.

`test_every_dropped_spec_is_named_in_the_header` proves the spec *appears* in the header; it does not prove the remedy is attached to it. Since the whole argument for keeping withdrawn and deprecated as separate categories is that the remedy differs, the remedy is the part worth asserting.

## 3. `test_generated_manifest_documents_the_excluded_side` is misnamed

`tests/unit/test_mcp_cli.py` — the test no longer asserts anything about the *excluded* (collision-loser) category; it now asserts the *withdrawn* one. The docstring explains the swap honestly, but the name misleads a reader scanning for collision coverage. Suggest `test_generated_manifest_documents_the_dropped_side`.

## 4. CHANGELOG: reconcile the counts

The #169 entry states, at various points, 28 / 21 / 19 / 8 / 7 / 4. These are all correct but measured at different layers:

- **28** dead endpoint *declarations* found by the live probe
- **21** deprecated with a replacement the library already ships
- **19** `WITHDRAWN_TOOLS` entries — MCP *tool keys*, not endpoints
- **4** of those 19 have no successor at all (`investment.mutual_fund_by_name`, `market.pre_post_market`, `institutional.asset_allocation`, `institutional.fail_to_deliver`)

Endpoint-level and MCP-tool-level counts legitimately differ, but a reader will try to reconcile them and fail. One sentence naming which layer each number counts would fix it.

## Provenance

All four raised during review of #169 and deliberately deferred to keep that PR's scope contained. #169 itself is green.

MehdiZare:
Closes #165
Closes #171

Also fixes the silent half of #166 (see below); **#166 itself stays open, targeted at 3.0.**

## #165 — one fact, stated twice

An endpoint declared parameters in `mandatory_params`/`optional_params`, and each param *also* carried a `required` flag saying the same thing. #155 made the two agree and guarded it; this removes the redundancy so they cannot disagree again.

`required` is now a **read-only property**, stamped from list membership at construction:

```python
required is property: True
p.required = True      -> AttributeError
```

The expression at the one public read site — `validate_value`'s `if self.required:` — is unchanged. Threading it in as `validate_value(value, required=...)` was rejected because it would stop `mandatory.validate_value(None)` raising, which is the exact public path #165 flags.

**Removed vs deprecated:** the 544 in-package `required=` declarations are gone (no external contract). The `required=` *constructor argument* is deprecated to 3.0 per the #136/#147/#174 idiom, and deliberately **kept in positional slot 4** — dropping it would silently shift a positional call and land `True` in `description`.

Disclosed cost: `description` gains a default, so it is no longer type-enforced until 3.0. A test covers what mypy no longer can.

## The silent half of #166

`ttl_overrides` is a public `dict[str, int]` keyed on endpoint name, read with a plain `.get(name, default)`. A typo'd or stale key was accepted silently and did nothing — no warning, no error, no way to notice except by watching request volume. Verified before the fix:

```python
CacheConfig(ttl_overrides={"totally-bogus-endpoint": 3600})   # accepted, does nothing
```

Now:

```
cache ttl_overrides names no known endpoint: 'serch-name' (closest endpoint: 'search-name')
```

This is a bug in its own right, independent of the rename. It is also what made #166 dangerous: renaming `search-name` would have broken users' TTL config invisibly. With this in place, the 3.0 rename is *announced* rather than silent.

**#166 stays open**, commented with the full analysis and a 3.0 recipe. Caches do invalidate — `_build_cache_key` uses the endpoint name as a literal unhashed prefix and the `file`/`redis` backends persist — which is major-version-appropriate, not minor.

## #171 — all four review follow-ups

Pair-keyed e2e exemption (zero collisions today, held by mypy plus a new no-network test), `test_withdrawn_drops_name_their_remedy`, the `_the_dropped_side` rename, and the count reconciliation. The #169 CHANGELOG entries contradicted themselves — "four" in one place, "six" in another, for the withdrawn endpoints with no successor. Four is correct; fixed.

## Verification

```
make lint / format / mypy   clean (both extras, incl. tests)
unit (extras)               1364 passed, 10 skipped
unit (core-only)            1171 passed, 17 skipped, 1 failed*
```

\* Pre-existing, reproduced on a detached worktree at `origin/dev` before any change: `test_the_spec_column_is_never_ellipsised_on_a_default_terminal` silently requires `rich`. Filed as **#180** and fixed in **#182**; this branch predates that fix, so merging #182 first clears it.

Mutation-tested: #165 guard 7/7 caught (including re-adding a setter, deleting the stamping, and a vacuity probe), TTL guard 4/4, withdrawn-header 3/3, exemption guard 1/1.

## Known limitation

`Endpoint` lacks `validate_assignment`, so reassigning `mandatory_params` after construction does not re-stamp requiredness. No code does this, and it was equally true of the stored flag before.


MehdiZare:
## Review follow-up

Independent code review — Important findings addressed in `4691242`:

1. **TTL key examples fixed** — README/docs/tests used `get_quote` / method-style keys; live lookup uses `Endpoint.name` (`quote`, `profile`). Official examples and fixtures now match.
2. **Warning remediation text** no longer points at `fmp-mcp list` (MCP tool names ≠ Endpoint.name); points at `docs/api/endpoints.md` / endpoint registry. Pluralized for multi-key lists.
3. **Positional slot 4 BC test** — `EndpointParam(name, loc, type, True, "desc")` still warns, keeps description, detached `required is True`.
4. Documented known limitation: no re-stamp of requiredness after post-construction param-list mutation.

**Deferred:** full #166 endpoint rename remains 3.0 (this PR only the warn-half). CI `claude-review` failure is #184.

Tests: param-required + cache TTL/backends/integration — 69 passed.

PR #185, opened 2026-08-09
fix(ci): soft-fail Claude Code Review when auth is missing (#184)

MehdiZare:
Closes #184

## Problem

The `claude-review` job failed every PR targeting `dev` without producing a review:

- Claude SDK returned `is_error: true`, `total_cost_usd: 0`, `num_turns: 1` in ~400ms
- Action log showed empty `ANTHROPIC_API_KEY`; the step uses `CLAUDE_CODE_OAUTH_TOKEN`
- No inline comments were posted
- The check is **not** required by the branch ruleset, but it still painted CI red

## Fix

1. **Skip when unconfigured** — job `if` requires `secrets.CLAUDE_CODE_OAUTH_TOKEN != ''` (in addition to same-repo)
2. **Fail soft when expired** — `continue-on-error: true` on the review step so a present-but-invalid token cannot fail the workflow

Review remains advisory. Required checks stay in `ci.yml`.

## Note

Restoring a valid `CLAUDE_CODE_OAUTH_TOKEN` in repo secrets is still the way to get actual reviews again; this PR only stops the red noise while auth is broken.

Issue #184, opened 2026-08-09
ci: claude-review fails with empty Claude auth (is_error:true, 0 cost)


MehdiZare:
## Problem

The `claude-review` job on PRs targeting `dev` fails consistently without producing a review:

- Workflow: `.github/workflows/claude-code-review.yml`
- Symptom: Claude SDK returns `subtype: success` with `is_error: true`, `total_cost_usd: 0`, `num_turns: 1`, in ~400ms
- Action log shows `ANTHROPIC_API_KEY` empty; the step uses `claude_code_oauth_token: ${{ secrets.CLAUDE_CODE_OAUTH_TOKEN }}`
- No inline comments are posted

Seen on PRs #181, #182, #183 (and likely any PR while the secret is missing/expired).

## Impact

- Red check on every PR (noise)
- No automated code review from the Claude action
- Not a required status check for `dev` (ruleset only requires `tests (3.10–3.14)`), so it does not block merge by ruleset, but it still marks CI as failed and confuses merge readiness

## Expected fix options (pick one)

1. **Restore auth**: ensure `CLAUDE_CODE_OAUTH_TOKEN` (or equivalent) is present and valid in repo secrets, re-run the job
2. **Fail soft**: set `continue-on-error: true` on the review step so missing credentials do not fail the PR
3. **Gate the job**: skip when the secret is unset, or only run when configured

## Out of scope for feature PRs

Do not fix this inside unrelated product PRs; it is infrastructure.

## Related

- CodeRabbit auto-reviews are also skipped for non-default base branches (`dev`); that is separate configuration in `.coderabbit.yaml` / CodeRabbit UI.

MehdiZare:
Fixed by #185 (merged to dev).

PR #186, opened 2026-08-09
fix(lc): dispatch tools through client methods, not bare request (#172)

### LINKED ISSUE #172: LangChain tools dispatch through client.request, bypassing client-method defaults
## What

The LangChain tool layer dispatches through `client.request(ENDPOINT, **kwargs)` rather than through the corresponding client method. The client methods carry behaviour the endpoint declaration does not.

## Why it matters, concretely

#169 made `from`/`to` **mandatory** on `sec.sec_filings_search_symbol`, `_cik` and `_form_type`, because the live API answers `400 Invalid or missing query parameter - from` without them.

That change is safe for every current caller **only because** each client method defaults the window to the last 30 days (`fmp_data/sec/client.py:148-149`), so the mandatory params are always populated before `validate_params` runs.

A LangChain tool that bypasses the method and calls `client.request(SEC_FILINGS_SEARCH_SYMBOL, symbol=...)` gets no such default. It now raises a local `ValidationError` where it previously received a remote `400`.

Both are errors, so **this is not a regression** — nothing that worked has stopped working. But it demonstrates the general gap: any behaviour a client method adds on top of its endpoint (defaults, one-of constraint enforcement, pagination, post-processing) is invisible to the LC surface, and the two can diverge silently.

`SECClient.search_industry_classification` is a second instance: it enforces the "at least one of `cik`/`sicCode`/`symbol`" constraint that `mandatory_params` cannot express. An LC tool calling the endpoint directly does not enforce it.

## Direction

Have the LC tool layer dispatch through the client method named in `EndpointSemantics.method_name` — which it already records — instead of reconstructing the call from the `Endpoint`. That makes method-level behaviour the single source of truth for both surfaces.

Needs design: the method signature and the endpoint parameter set are not always the same shape, so the argument mapping is the substance of the work.

## Provenance

Raised during review of #169 and flagged there as out of scope. #169 already notes it as unaddressed.

MehdiZare:
Closes #172

## Problem

LangChain tools built by `EndpointVectorStore.create_tool` dispatched through `client.request(endpoint, **kwargs)`, skipping every behaviour the matching client method adds:

- Default date windows (SEC filings search methods default `from`/`to` to the last 30 days)
- One-of constraints (`search_industry_classification` requires at least one of `symbol`/`cik`/`sic_code`)
- Pagination defaults and any post-processing

#169 made `from`/`to` mandatory on the SEC filings-search endpoints because the live API requires them. That is safe for direct client callers only because the methods fill the window. An LLM tool that only had a symbol raised a local `ValidationError` instead.

MCP already dispatches through `client.<client>.<method>` (`tool_loader._resolve_attr`). LC did not.

## Fix

1. **Resolve** `client.<semantics.client_name>.<semantics.method_name>`
2. **Map** wire/tool kwargs onto the method signature (`from`→`from_date`/`start_date`, `periodLength`→`period_length`, `sicCode`→`sic_code`, …)
3. **Reclassify** schema optionality from method defaults — endpoint-mandatory params the method defaults become optional so the LLM may omit them
4. **Surface** method-level `ValueError` as structured `validation_error`
5. **Fall back** to `client.request` when the store holds a bare `BaseClient` / test double without sub-clients (pre-#172 behaviour)

Tool field *names* stay the endpoint/wire names for continuity; only requiredness and dispatch change.

## Tests

`tests/unit/lc/test_method_dispatch.py` covers alias mapping, partition, SEC symbol-only invoke (no `request`), request fallback, and industry-classification one-of.

## Verification

- `uv run pytest tests/unit/lc/ -q` → 77 passed
- pre-commit (ruff, mypy) clean


MehdiZare:
## Review + fixes (post-push)

Full review of #186 found one **critical** regression risk and two important polish items. Isolated fixes are in `d9afa0a`:

### Fixed in this PR
1. **Compatibility gate** — if a client method has a required parameter that no endpoint/wire field can supply (e.g. Form 13F / institutional holdings: wire `year`/`quarter` vs method `report_date`), dispatch falls back to `client.request` + the pre-#172 schema instead of `TypeError` at invoke.
2. **`trade_date` alias** — endpoint `date` maps to method `trade_date` (`get_insider_trading_latest`).
3. **Schema honesty under method dispatch** — unmapped endpoint params are omitted from the tool schema when method dispatch is active (so LLMs are not asked for values that would be silently dropped).
4. Tests cover the gate, Form 13F fallback, `trade_date`, and omit-unmapped behaviour.

### Follow-up issues (not expanded here)
- #187 — restore Claude OAuth + soft-fail visibility (ops for the advisory reviewer)
- #188 — shared LC/MCP tool-binding layer, catalogue-wide endpoint↔method guard, institutional year/quarter vs `report_date` shape debt

### Notes
- Claude review was red on earlier runs because of #184; fixed on `dev` via #185. Re-running CI against current `dev`.
- Scope kept tight: no year/quarter→date synthesis, no MCP rewrite.

PR #189, opened 2026-08-09
fix(lc): endpoint↔method coverage guard and tool-arg docs (#188)

MehdiZare:
## Summary

Addresses **#188** (isolated guard + docs). **#187** is closed as not planned (OAuth token dropped; no restore/rotate).

### #188 — endpoint↔method coverage guard + docs
- New `tests/unit/lc/test_endpoint_method_coverage.py` walks every `(endpoint_map × semantics)` pair and asserts:
  1. Every semantics `method_name` resolves on a live client
  2. Request-fallback set is exactly the two known institutional tools (`get_form_13f`, `get_institutional_holdings` — wire `year`/`quarter` vs method `report_date`)
  3. Mandatory wire fields dropped under method dispatch are exactly the two revenue `structure` params
- New mismatches fail CI until allowlisted (with a comment) or fixed
- README / `docs/mcp/index.md` / `LLM.md` document: LangChain tool args = API/wire names; MCP tool args = Python method parameter names

**Not in this PR (still open under #188):** shared LC/MCP tool-binding module, institutional year/quarter ↔ `report_date` shape alignment.

### #187 — not planned
OAuth token dropped; no soft-fail summary / rotate docs. Workflow still skips when the secret is empty (#184/#185).

### #166 — no change
Deferred to 3.0.

## Test plan
- [x] `uv run pytest tests/unit/lc/test_endpoint_method_coverage.py tests/unit/lc/test_method_dispatch.py -q` → 18 passed
- [x] pre-commit clean on commit
- [ ] CI green on this PR

Issue #188, opened 2026-08-09
lc/mcp: shared tool-binding layer and endpoint↔method param coverage guard


MehdiZare:
## Context

#172 / #186 teach LangChain tools to dispatch through `client.<client>.<method>` (the path MCP already uses), with wire→method param aliases and a compatibility fallback when shapes diverge.

### 1. Shared tool-binding module
MCP (`fmp_data/mcp/tool_loader._resolve_attr`) and LC (`fmp_data/lc/vector_store.resolve_*`) both resolve client methods. MCP exposes Python method signatures as tool args; LC keeps wire names and maps at invoke. A shared binding layer would reduce drift.

### 2. CI guard: endpoint↔method coverage
A unit/CI guard should walk `(endpoint map × client method)` pairs and fail when:
- method dispatch would `TypeError` (required method param uncovered), or
- a mandatory wire field is silently dropped under method dispatch

#186 adds a runtime compatibility gate; a catalogue-wide guard would catch new mismatches on every PR.

### 3. Endpoint vs method shape debt (institutional)
Several institutional tools expose wire `year`/`quarter` while client methods take `report_date` (e.g. `get_form_13f`, `get_institutional_holdings`). #186 correctly falls back to `client.request` for those. Long-term options:
- Align endpoint declarations / LC schemas with method shapes, or
- Add explicit bridge helpers (no silent year/quarter→date invention in the LC layer)

### 4. Docs
Document that LC tool arg names are API/wire names, and that MCP tools use Python method parameter names.

## Non-goals for a single PR
Do not rewrite MCP and LC together in one change. Prefer isolated follow-ups per bullet.

Related: #172, #186

MehdiZare:
Addressed the isolated, non-rewrite pieces in #189:

1. **CI guard** — `tests/unit/lc/test_endpoint_method_coverage.py` walks every `(endpoint_map × semantics)` pair; pins the two request-fallback institutional tools and the two dropped-mandatory revenue `structure` fields; any new mismatch fails CI.
2. **Docs** — README / `docs/mcp/index.md` / `LLM.md` state that LangChain keeps wire arg names and MCP exposes Python method parameter names.

Still open under this issue (as scoped):
- Shared LC/MCP tool-binding module
- Institutional `year`/`quarter` vs `report_date` shape alignment (or explicit bridge helpers)

MehdiZare:
Merged via #189 (squash `964c452`).

**Shipped in this slice**
[...]

PR #190, opened 2026-08-09
fix(institutional,lc): align 13F tools with the API's year/quarter shape (#188)

MehdiZare:
Closes the third bullet of #188 (institutional endpoint↔method shape debt). The shared LC/MCP tool-binding module — bullet 1 — follows in a separate PR, per the issue's "prefer isolated follow-ups per bullet".

## Why

The last two request-fallback tools in the LangChain catalogue existed because their client methods required a `report_date` the wire does not have. #188 offered two directions — align the shapes, or add explicit bridge helpers. Probing the live API settles which side was wrong:

```
GET /stable/institutional-ownership/extract?cik=0001067983
-> 400  Query Error: Invalid or missing query parameter - year

GET /stable/institutional-ownership/symbol-positions-summary?symbol=AAPL
-> 400  Query Error: Invalid or missing query parameter - year
```

`year` and `quarter` are mandatory on both endpoints, neither accepts a date parameter at all, and the period-end `date` is a **response** field. `report_date` was a client-side invention; the endpoint declarations and the semantics `parameter_hints` were already right.

## What

- New wire-shaped methods on both the sync and async institutional clients:
  `get_form_13f_by_quarter(cik, year, quarter)` and `get_institutional_holdings_by_quarter(symbol, year, quarter)`.
- `get_form_13f(cik, report_date)` and `get_institutional_holdings(symbol, report_date, ...)` **keep their signatures** and delegate. The request call, error swallowing and empty-result logging move onto the wire-shaped methods, so the tool layers get them too.
- `form_13f` / `institutional_holdings` semantics `method_name` (and the matching endpoint-map keys) point at the new methods.

## Tool-surface effects

| | before | after |
|---|---|---|
| MCP tool name | `form_13f` | unchanged |
| MCP tool args | `cik`, `report_date` | `cik`, `year`, `quarter` |
| LangChain tool name | `get_form_13f` | `get_form_13f_by_quarter` |
| LangChain tool args | `cik`, `year`, `quarter` | unchanged |

The MCP args now match the `parameter_hints` those semantics entries already declared, and an assistant no longer has to invent a quarter-end date to ask for a quarter. The LangChain rename follows from tool names coming from `EndpointSemantics.method_name`. The Python API is untouched.

## Guard

`KNOWN_REQUEST_FALLBACK` in the #189 coverage guard is now **empty** — every tool in the catalogue dispatches through a client method. The allowlist and its assertions stay rather than being deleted with the entries: an empty allowlist is the strongest form of the guard. A new synthetic test keeps the request-fallback branch of `create_tool` covered now that no real endpoint exercises it.

Also fixes `test_endpoint_group_organization`, which reimplemented one of the three rules in `resolve_semantics_for_endpoint` (strip a leading `get_`) rather than calling it. Keys resolving by exact match or by `method_name` read as *missing* there while registering fine in the real registry — a guard that reimplements the rule it guards fails on correct code and would pass on code the resolver rejects.

## Verification

- 1425 unit tests pass with `uv sync --all-extras`; 1196 pass with plain `uv sync` (CI installs no extras).
- `ruff check`, `ruff format --check`, and `mypy fmp_data tests` clean.
- Both new methods called against the live API: `get_form_13f_by_quarter("0001067983", 2023, 3)` returns 46 holdings, the date-shaped wrapper returns the same 46 for any date inside Q3, and `get_institutional_holdings_by_quarter("AAPL", 2023, 3)` returns the AAPL summary with `date=2023-09-30`.
- MCP signatures confirmed through the real loader: both tools now advertise `(cik|symbol, year: int, quarter: int)`.


MehdiZare:
@coderabbitai review

MehdiZare:
## Review pass (worktree) — findings addressed

Full code review of this PR (no CodeRabbit auto-review on non-default base; rate-limited on re-trigger).

### Addressed in 7cbe50a
1. **CHANGELOG BREAKING labels** for MCP arg change (`report_date` → `year`/`quarter`) and LangChain tool renames (`get_*_by_quarter`), plus a migration one-liner.
2. **Documented LangChain Form 13F envelope change** — method dispatch means API/validation errors now surface as success+empty (matching the Python method), not `status: error` from the old request-fallback path.
3. **Stale request-fallback docstring** rewritten for the empty allowlist.
4. **Symmetric LC dispatch test** for `get_institutional_holdings_by_quarter`.

### Deferred to isolated follow-ups (not bloating this PR)
- #192 remaining institutional year/quarter wrappers
- #193 async exception narrowing + empty-on-error policy

No critical correctness issues. Python API signatures unchanged; wire shape alignment is sound.

PR #191, opened 2026-08-09
fix(lc,mcp): one shared endpoint→method binding layer (#188)

MehdiZare:
Closes the first bullet of #188 (shared tool-binding module). Together with #190 this finishes the issue's remaining scope.

> **Stacked on #190.** Base is `fix/institutional-wire-shape-alignment` so the diff here shows only this change. Merge #190 first; I'll retarget this to `dev` once it lands.

## Why

MCP and LangChain answer the same question — *given an endpoint and the semantics naming its client method, how do I call that method?* — and both answered it separately. `mcp.tool_loader._resolve_attr` walked the attribute chain; `lc.vector_store` kept the alias table, the coverage gate and the invoke-time kwargs mapping. #188 calls this drift risk.

It was not hypothetical. The "which parameters of this method can receive a value" filter existed in **four** hand-written copies — three in `vector_store`, one more in the catalogue guard — and they disagreed:

```python
# method_dispatch_compatible / map_tool_kwargs_to_method
if name != "self" and param.kind in (POSITIONAL_OR_KEYWORD, KEYWORD_ONLY)

# partition_params_for_method
if name != "self"
```

So for `def method(symbol, **kwargs)`, `kwargs` was a real parameter to one copy and not the other — and since a `VAR_KEYWORD` parameter has no default, it read as a *required* parameter named `kwargs`.

## What

New `fmp_data/tool_binding.py`, importing nothing but `inspect`. It sits in the core package rather than under `lc/` or `mcp/` on purpose: a shared layer reachable only with an optional extra installed would not be shared.

- **`bindable_params()`** — the one definition of a fillable parameter. The surviving behaviour is the stricter one (`*args`/`**kwargs` excluded).
- **`uncovered_required_params()`** — names the required method params no wire field can fill. `method_dispatch_compatible()` is now defined as its absence, so the two cannot disagree.
- **`resolve_attr()` (raises) vs `resolve_client_method()` (returns `None`)** — documented as two reporting styles over one walk, not an inconsistency. MCP must refuse to start on a ghost `method_name`; LangChain must fall back to `client.request` on a client with no sub-clients.
- Plus the alias table, `method_param_aliases`, `resolve_method_param_name`, `map_tool_kwargs_to_method`, `partition_params_for_method`, `resolve_dispatch_method`.

`tests/unit/lc/test_endpoint_method_coverage.py` now imports the shared helpers instead of keeping private copies of them — a guard that reimplements the rule it guards can only check itself.

## Compatibility

Every name previously importable from `fmp_data.lc.vector_store` still is, including the private `_camel_to_snake` and `_ENDPOINT_TO_METHOD_ALIASES`, and `fmp_data.mcp.tool_loader._resolve_attr`. No caller changes. No behaviour changes except the `**kwargs` classification above, which no client method in the catalogue hits.

## Verification

- 1442 unit tests pass with `uv sync --all-extras`.
- 1213 pass with plain `uv sync`, **up from 1196**: `tests/unit/test_tool_binding.py` deliberately lives outside `tests/unit/lc/` so it runs in the core-only environment CI actually installs — which is the environment the module exists to work in.
- `ruff check`, `ruff format --check`, `mypy fmp_data tests` clean.
- Four mutations of the new module, each killed by a named test:
  | mutation | test that caught it |
  |---|---|
  | `bindable_params` loses the kind filter | `test_var_keyword_does_not_make_a_method_look_required` |
  | `resolve_attr` stops rejecting non-callables | `test_strict_resolver_rejects_a_non_callable` |
  | alias precedence reversed | `test_aliases_prefer_exact_match_over_rename` |
  | `map_tool_kwargs_to_method` keeps `None` | `test_map_tool_kwargs_renames_and_drops` |


MehdiZare:
@coderabbitai review

MehdiZare:
## Review pass (worktree) — findings addressed

Stacked on #190. Full code review of the shared binding extraction.

### Addressed in 0df93ee (rebased onto #190)
1. **`resolve_client_method` now wraps `resolve_attr`** — true "one walk, two reporting styles"; cannot drift on edge cases.
2. **`SearchResult` restored** on `vector_store.__all__` (star-import parity).
3. **Re-export identity pins** for MCP `_resolve_attr` and LC private/public aliases + `SearchResult` in `__all__`.
4. **CHANGELOG Fixed bullet** for the `**kwargs` / `bindable_params` unification.

### Deferred to isolated follow-ups
- #194 warn on silent LangChain request-fallback at tool creation
- #195 POSITIONAL_ONLY/`cls`, multi-alias conflicts, VAR_KEYWORD passthrough

No critical bugs. Architectural placement in core is correct; catalogue still method-dispatches everywhere.

Issue #195, opened 2026-08-09
tool_binding: POSITIONAL_ONLY/cls, multi-alias conflicts, VAR_KEYWORD passthrough


MehdiZare:
## Context

Follow-ups from #191 that do not block the shared binding extraction:

1. \`bindable_params\` only drops \`self\`, not \`cls\` on classmethods; \`POSITIONAL_ONLY\` required params are excluded so dispatch can look compatible then \`TypeError\` at invoke.
2. \`map_tool_kwargs_to_method\` last-wins when two wire aliases map to the same method param (e.g. \`from\` and \`from_date\`) with no conflict signal.
3. Methods with \`**kwargs\` still drop unmapped wire names (they never enter \`**kwargs\`). Fine for current catalog; document or guard against future passthrough methods.

No current catalog methods hit these. Optional unit cases only.

Related: #191

MehdiZare:
Landed in #199.

Issue #194, opened 2026-08-09
lc: warn when method dispatch falls back to client.request at tool creation


MehdiZare:
## Context

From review of #191: when \`resolve_dispatch_method\` returns \`None\` (missing sub-client, missing method, or unfillable shape), LangChain silently falls back to \`client.request\` with no log/warning. Correct for bare \`BaseClient\` / tests; easy to miss in a misconfigured production store.

The catalogue guard only covers the live catalog, not a runtime store whose semantics drifted.

## Proposal

At tool creation, log WARNING when falling back, distinguishing:

- no sub-client present (expected for bare clients)
- sub-client present but method missing / shape mismatch (likely misconfiguration)

Related: #188, #191

MehdiZare:
Landed in #199.

Issue #193, opened 2026-08-09
institutional: narrow async Form 13F exception swallowing + empty-on-error policy


MehdiZare:
## Context

From review of #190:

1. **Sync/async asymmetry:** \`get_form_13f_by_quarter\` (sync) catches \`(FMPError, ValidationError)\` and returns \`[]\`; the async counterpart still catches bare \`Exception\`, which can hide programming bugs as "no 13F data."
2. **Empty-on-error policy:** LangChain Form 13F now returns \`status: success\` + empty data on API errors (because it dispatches through the method). Decide whether empty-on-error remains the long-term contract for 13F, or whether tools should surface errors.

## Proposal

- Align async to \`(FMPError, ValidationError)\` (and same for \`get_form_13f_dates\`).
- Document or revise the empty-on-error convenience so agent surfaces stay predictable.

Related: #190

MehdiZare:
Landed in #197.

PR #196, opened 2026-08-09
fix: post-#188 review follow-ups (#192, #193, #194, #195)

### LINKED ISSUE #192: institutional: align remaining year/quarter wrappers to wire shape
## Context

#190 added wire-shaped `get_form_13f_by_quarter` / `get_institutional_holdings_by_quarter` for the two catalogue tools that still fell back to `client.request`. Sibling institutional methods still invent a client-side `report_date` and convert to year/quarter internally:

- `get_institutional_ownership_extract` (same wire path as FORM_13F)
- `get_symbol_positions_summary` (same wire path as INSTITUTIONAL_HOLDINGS)
- analytics / industry breakdown / performance helpers that take `report_date`

## Proposal

Add matching `*_by_quarter` (or shared helpers) for the remaining year/quarter endpoints, keep date-shaped wrappers as thin delegates, and decide whether any of them should become tool methods.

## Non-goals for this issue

Do not expand #190 / #191 scope; those PRs intentionally isolated the two catalogue tools.

Related: #188, #190

MehdiZare:
Closes #192, #193, #194, #195 — the four issues filed from review of #190/#191. The binding-layer half and the institutional half touch disjoint code but share one origin, so they land together.

## #195 — a positional-only required param passed the gate, then crashed the call

`bindable_params` lists parameters fillable **by name**, and dispatch is `method(**mapped)`. A required `POSITIONAL_ONLY` parameter was merely *absent* from that map, which read as "nothing required is missing". Reproduced on `dev` before the fix:

```python
def f(cik, /, year=2020): ...
method_dispatch_compatible(f, ["cik", "year"])          # True
map_tool_kwargs_to_method(f, {"cik": "…", "year": 2023}) # {'year': 2023}  <- cik dropped
f(**mapped)  # TypeError: f() missing 1 required positional argument: 'cik'
```

That is exactly the failure the gate exists to prevent. New `keyword_unfillable_required_params()` reports them; `uncovered_required_params()` includes them, so such a method falls back to `client.request`. No catalog method has one today.

Also in #195:
- **`cls` is dropped alongside `self`.** A bound classmethod hides its receiver, but the catalogue guard and the unit tests pass the underlying function, where `cls` read as a required parameter no wire field could fill.
- **Conflicting aliases raise instead of last-wins.** If `from` and `from_date` both resolve to `from_date` with *different* values, silently keeping whichever iterated last answers about the wrong date window. Now a `ValueError` naming both keys. Agreeing values are not a conflict. Unreachable from a well-formed tool schema (`extra="forbid"`), so this protects direct callers.
- **`**kwargs` non-passthrough** is documented and pinned by a test rather than changed — a future passthrough method should opt in, not inherit whatever this happens to do.

## #194 — the fallback is announced now

`MethodBinding` / `bind_client_method` report *why* there is no method: `NO_SUBCLIENT`, `NO_METHOD`, or `SHAPE_MISMATCH` (carrying the uncovered params). `create_tool` logs:

- **DEBUG** for no sub-client — normal for a bare `BaseClient`, and warning would emit one line per endpoint in the catalogue, burying the two that matter.
- **WARNING** naming the cause for the other two, which mean something is genuinely misconfigured.

This is a deliberate refinement of the issue's "log WARNING when falling back": the level *is* the distinction the issue asked for. `resolve_dispatch_method` keeps its `Callable | None` signature.

## #193 — async institutional exception narrowing

`get_form_13f_by_quarter` and `get_form_13f_dates` caught bare `Exception` and reported "no 13F data", so a `TypeError` in this library surfaced as an empty portfolio. Narrowed to `(FMPError, ValidationError)`, matching sync.

On the policy half: empty-on-error is **kept** — callers depend on it and narrowing it is breaking — and is now documented where callers meet it, including that the LangChain tool inherits it and answers `status: success` with empty data. Flagged for 3.0.

## #192 — remaining year/quarter wrappers, and one that does not filter

Five `*_by_quarter` siblings (extract, analytics, holder industry breakdown, symbol positions, industry performance), sync and async, with the date-shaped methods kept as thin delegates. Each verified live to actually honour the filter — 2023 Q3 → 2023-09-30 rows, 2019 Q1 → 2019-03-31 rows. None of the five is an MCP or LangChain tool (the issue asks: they are not in `INSTITUTIONAL_ENDPOINT_MAP`), so this is Python-API consistency only and no tool surface changes.

**The sixth is different, and it is the find of the batch.** `holder-performance-summary` *ignores* `year`/`quarter`:

```
cik=0001067983                      -> 52 rows, 2013-06-30 … 2026-03-31
cik=0001067983 & year=2023&quarter=3 -> byte-identical to unfiltered
cik=0001067983 & year=2019&quarter=1 -> byte-identical to unfiltered
```

So `get_holder_performance_summary(cik, report_date=...)` has been returning the holder's entire history while looking like it selected a period. It now emits a `UserWarning` saying so and pointing at the `date` field on each row, still sends the params in case FMP starts honouring them, and deliberately gains **no** `_by_quarter` sibling — advertising a filter the API does not apply would hide the problem rather than expose it.

## Verification

- 1469 unit tests pass with `uv sync --all-extras`; 1234 with plain `uv sync` (CI installs no extras).
- `ruff check`, `ruff format --check`, `mypy fmp_data tests` clean. The new fallback logging pushed `create_tool` over the complexity limit, so it was extracted to `_log_dispatch_fallback` rather than suppressed.
- The `POSITIONAL_ONLY` crash reproduced against `dev` before the fix and covered by a test after.
- Sync/async signature parity asserted for all seven `*_by_quarter` methods.
- The five new methods called against the live API; all return 2023-09-30 rows for 2023 Q3.


MehdiZare:
Superseded by the isolated split:

- #197 — #193 only (async exception narrowing)
- #198 — #192 only (remaining institutional \`*_by_quarter\`)
- #199 — #194 + #195 (tool dispatch hardening)

Closing this mega-PR to avoid double-landing. Prefer the split for independent review/merge.

MehdiZare:
Closing in favor of isolated PRs #197, #198, #199.

PR #205, opened 2026-08-10
fix(ci): stop silent successes in the release path (#202, #203, #204)

### LINKED ISSUE #202: ci: dev and main diverge after every release, and the next release PR silently gets no CI
## What happens

Every release squash-merges `dev` → `main`, which puts a commit on `main` that is not an ancestor of `dev`. From then on the two branches share no recent history. Nothing notices until the *next* release PR, when git takes the merge base back to the previous release era and `dev → main` comes out **CONFLICTING**.

Hit while cutting 2.6.0: PR #201 opened with 23 conflicting files.

## Why it is worse than a normal conflict

**A conflicting PR gets no CI at all.** GitHub cannot build a merge ref, so *no* `pull_request`-triggered workflow runs — Test-Matrix, `document.yml`, `guard-main-origin` were all simply absent from the checks list. Only CodeQL reported, because it runs on a different event.

The release PR therefore looks *quiet*, not red. Nothing says "conflicted"; the checks list just has a hole in it where the test matrix should be. It would be easy to read that as "still queued" and wait, or to label and merge on the strength of the checks that *did* pass.

## Why `sync-main-to-dev` does not catch it

It is not misbehaving. It fires on push to `main` and compares **content**:

```yaml
DIFF=$(git diff origin/dev..main --stat)
```

Immediately after a release squash, `dev` and `main` have identical content — the squash was made from `dev` — so the diff is empty and it correctly does nothing. The *histories* are already divergent at that moment. They only drift into conflict later, as `dev` moves on.

So the check is asking the wrong question: content equality, when the problem is reachability.

## Fix applied for 2.6.0 (manual)

`eea540a` on `dev`: `git merge -s ours origin/main`, after verifying `dev` was already a content superset of `main` —

- `main` had exactly one commit not on `dev` (the 2.5.0 squash `3c754f8`);
- its tree `af5d5e6` is byte-identical to dev's `d9a37cc`;
- no file exists on `main` that is absent from `dev`.

Tree hash `d5c45c4` before and after, so no content changed. `dev → main` went straight from CONFLICTING to MERGEABLE and the full matrix ran.

## Proposed

1. **Change the sync trigger from content to reachability.** Replace the `git diff` guard with `git merge-base --is-ancestor origin/main origin/dev`; if false, open (or push) a back-merge into `dev`. That fires at exactly the right moment — right after the release, when the fix is a trivial no-content merge — instead of at the start of the next release, when it is 23 conflicts.
2. **Fail loudly when a release PR has no merge ref.** A required check that asserts `mergeable != CONFLICTING` would turn a silent gap in the checks list into a red X. Without it, "no CI ran" is indistinguishable from "CI has not started yet".

Related: #203 (the `Release-PR` workflow reports success while creating nothing).

### LINKED ISSUE #203: ci: the Release-PR workflow can never create a release PR, but reports success
## What happens

`.github/workflows/release-pr.yml` runs on every push to `dev` and is meant to open the `release/dev-to-main` PR. It has **never** created one. It exits green every time.

Its own log, from the run on `70dbb5f`:

```
Working base is branch 'dev'
Pull request branch to create or update set to 'release/dev-to-main'
Rebasing commits made to branch 'dev' on to base branch 'main'
Pull request branch 'release/dev-to-main' does not exist yet.
Switched to a new branch 'release/dev-to-main'
Branch 'release/dev-to-main' is not ahead of base 'main' and will not be created
```

## Cause

`peter-evans/create-pull-request` is designed to commit **uncommitted working-tree changes** onto a new branch and open a PR for them. The job checks out `dev` and then runs the action without modifying anything, so there is nothing for it to commit: it resets onto `main`, finds the branch is not ahead, and stops.

It is the wrong tool for this job. Opening a PR between two branches that already exist needs `gh pr create --base main --head dev`, not a commit-and-PR action.

## Why it matters

The workflow reports **success**, so the failure is invisible. The documented release process is "push to dev → release PR appears → label it", and step two silently does not happen. Both 2.5.0 (#113) and 2.6.0 (#201) were in fact released through hand-made `dev → main` PRs; the automation contributed nothing while appearing to work.

## Proposed

Replace the action with a direct create-if-absent, e.g.:

```yaml
- name: Create or update the release PR
  env:
    GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
  run: |
    if gh pr list --base main --head dev --state open --json number \
         --jq '.[].number' | grep -q .; then
      echo "Release PR already open; nothing to do."
    else
      gh pr create --base main --head dev \
        --title "Release: dev → main" --body-file .github/release-pr-body.md
    fi
```

Note this only works once #202 is fixed — while `dev` and `main` have divergent histories the PR is CONFLICTING, and a conflicting PR gets no CI at all.

Alternatively, delete the workflow and document that the release PR is opened by hand. Either is better than a green check for work that did not happen.

### LINKED ISSUE #204: ci: TestPyPI validation silently skips every build after a PR's first, but reports success
## What happens

`publish-testpypi.yml` derives the TestPyPI version from the **PR number**:

```bash
TEST_VERSION="${NEW_VERSION}.dev${{ github.event.pull_request.number }}"
```

That is constant across every push to the same PR. The upload step then runs with:

```yaml
- uses: pypa/gh-action-pypi-publish@v1.14.2
  with:
    skip-existing: true
```

So the first push to a release PR publishes `X.Y.Z.devNNN`, and **every later push re-uploads the same version, gets skipped, and the job still reports success.** TestPyPI keeps serving the first build forever.

## Why it matters

The job comments on the PR with:

> ✅ Once verified, merge this PR to trigger the actual release to PyPI.

Anyone who follows that instruction after a fix installs **stale code** and concludes the release is good.

Observed on #201 while cutting 2.6.0. After pushing `a5badb4` (two real fixes), the workflow re-ran green and the comment still advertised `2.6.0.dev201`. Installing that version gave the pre-fix build:

```
dist metadata: 2.6.0.dev201
__version__  : 0.0.0          <- the fix in a5badb4 was not there
warning helper present: False <- nor the other fix
```

Both fixes were present in a wheel built locally from the same commit, so the artifact on TestPyPI was simply the older one.

This is the same failure shape as #202 and #203: **a green check for work that did not happen.** Three of them in one release cycle.

## Proposed

Make the test version unique per commit so `skip-existing` stops masking re-runs:

```bash
TEST_VERSION="${NEW_VERSION}.dev${{ github.event.pull_request.number }}+g$(git rev-parse --short HEAD)"
```

TestPyPI rejects local version identifiers (`+…`), so use a monotonic counter instead — e.g. the run number, which increases on every re-run:

```bash
TEST_VERSION="${NEW_VERSION}.dev${{ github.run_number }}"
```

and have the PR comment state the commit SHA it was built from, so a reader can tell whether the artifact matches the code they are reviewing.

Optionally drop `skip-existing: true`: with a unique version there is nothing legitimate to skip, and a collision then fails loudly instead of silently.

## Workaround used for 2.6.0

Verified the release candidate from a locally built wheel instead:

```
dist metadata : 2.5.1.dev44+ga5badb483
__version__   : 2.5.1.dev44+ga5badb483   (matches)
#166          : search_name / search-name
#192 warning  : fires on the by_quarter path, silent on the supported path
```

MehdiZare:
## Summary

Closes the three open CI issues that all share the same failure shape: **a green check for work that did not happen.** Observed together while cutting 2.6.0. Also includes a full CI audit pass for the same failure class in adjacent workflows.

### Original issues (#202 / #203 / #204)

| Issue | Bug | Fix |
|---|---|---|
| #202 | `Sync-Main-to-Dev` compared content (`git diff`), so post-squash history divergence was invisible until the *next* release PR opened CONFLICTING with no merge-ref CI | Key on `git merge-base --is-ancestor origin/main origin/dev`; open a reachability PR (`merge -s ours` when trees match) |
| #203 | `Release-PR` used `peter-evans/create-pull-request` with no working-tree changes → always green, never creates a PR | `gh pr create --base main --head dev`; fail if histories diverged |
| #204 | TestPyPI version = `X.Y.Z.dev{PR#}` + `skip-existing` → first build wins forever | Unique `dev{run_id}{run_attempt}` version; `skip-existing: false`; comment includes commit SHA |

Also on that pass: `Guard-Main-Origin` fails on `mergeable_state=dirty`.

### Audit pass (additional defects)

| Workflow | Bug | Fix |
|---|---|---|
| `claude-code-review.yml` | `secrets.*` in job-level `if` is illegal → failed workflow-validation check named after the file on every push that touches it; #184 soft-fail never evaluated the secret | Gate token presence in a step via env |
| `dev-release.yml` | Same #204 shape: `dev{commit_count}` + `skip-existing` → re-runs advertise stale wheels | Unique `dev{run_id}{run_attempt}`; `skip-existing: false` |
| `release.yml` | Existing tag skipped creation, job continued, PyPI `skip-existing: true` made collisions green; empty version possible; PR title interpolated into shell | Fail on existing tag/release/PyPI version; fail on empty version; checkout `merge_commit_sha`; env-pass untrusted fields |
| `rebase-reminder.yml` | New comment on every `synchronize` (spam) | `body-includes` + `edit-mode: replace`; delete when caught up |
| `ci.yml` secret-scan | Claimed "new findings only" via `--baseline` but never compared hashes | Drop baseline; cassettes are zero-tolerance |
| `document.yml` | `pages:write` + `id-token:write` on every PR build | Deploy job only |

### Left as follow-ups (ruleset / process, not workflow YAML)

- **Protect Main** required checks are only `tests (3.10–3.14)` + `CodeRabbit`. `Guard-Main-Origin`, `Test-MatrixExpected`, coverage, and secret-scan are **not** required — a conflicting release PR can still look quiet if those aren't required.
- **Protect Dev** requires signed commits and the five test matrix jobs (strict).
- **Main is not ruleset-protected for `Test-MatrixExpected`**, so consider requiring the aggregator.
- PRs opened by `GITHUB_TOKEN` (Release-PR / Sync) do not re-trigger `pull_request` workflows until the next human push or close/reopen.

## Test plan

- [x] `actionlint` clean on all workflows
- [ ] PR #205 CI green (Test-Matrix)
- [ ] After merge to `dev`, next push opens/reuses a real `dev → main` release PR
- [ ] After next release squash to `main`, Sync opens a history-reachability PR
- [ ] Labeled release PR: each push publishes a new TestPyPI version with matching commit SHA
- [ ] Claude Code Review no longer produces a failed "`.github/workflows/claude-code-review.yml`" validation check on push
- [ ] Rebase-reminder updates one comment rather than stacking

Closes #202
Closes #203
Closes #204


MehdiZare:
## Review pass (a51dc97)

Full review of #205 (workflow silent-success path + audit commit). **No unresolved human review threads** at start; CodeRabbit auto-review is skipped on non-default base (`dev`) — `@coderabbitai review` requested separately.

### Addressed in this push

| Finding | Fix |
|---|---|
| Guard-Main-Origin went **green** when `mergeable_state` stayed `unknown` after retries | Fail closed with `::error::` + `exit 1` |
| Sync conflict path force-pushed an **empty** tip / could wipe a good open reachability PR | Abort only; print recovery steps to log + step summary; **do not** call `open_or_update_pr` |
| `gh pr list … \|\| true` swallowed API/auth failures | List failures fail the job |
| GITHUB_TOKEN-created PR CI gap only in step summary | Documented in release PR body + `releasing.md` happy path |
| TestPyPI / Dev Release soft version grep | Assert sdist `Version:` matches expected before upload |
| `run_id`∥`run_attempt` concatenation collision class | Single `.devN` with `N = run_id * 1000 + run_attempt` |
| Outdated `skip-existing: true` example in releasing.md | Removed / replaced with pointer to live `release.yml` |

### Follow-up issues (broader scope, intentionally out of this PR)

- #206 — workflow-triggering token/App for automation-created PRs
- #207 — re-check open release PR health on every `dev` push
- #208 — sync conflict UX + concurrency hardening

`actionlint` clean on the touched workflows.

MehdiZare:
@coderabbitai review

MehdiZare:
**Merged** as `f4d54c2` (squash, GitHub-signed) into `dev`.

- All required CI green (tests 3.10–3.14 + adjacent checks)
- Review findings addressed in `a51dc97` before merge
- Zero unresolved review threads
- #202 / #203 / #204 closed manually (auto-close does not fire for non-default base)
- Broader follow-ups: #206, #207, #208

Issue #218, opened 2026-08-10
ci: document Guard base-pin lag for check-pr-mergeable contract changes


MehdiZare:
## Context

`Guard-Main-Origin` overlays `.github/actions/check-pr-mergeable` from `origin/<base>` when present (hotfix-safe pin from #210). Contract tightenings (e.g. #213 / #214 requiring `mergeable=true`) therefore apply to Guard only after the change ships to the PR base (typically `main` for release PRs). Release-PR on `dev` uses the tip workflow immediately.

## Proposed

- Document this lag in `docs/contributing/releasing.md` (and/or action.yml) so operators know when Guard semantics flip.
- Optionally evaluate whether the pin should allow forward-compatible minor contract updates (careful security/hotfix tradeoff — decision issue, not a silent code change).

## Out of scope

- Changing the pin strategy without an explicit decision

Related: #210, #213, #214

MehdiZare:
Addressed and merged via the corresponding PR (squash merge did not auto-close; closing explicitly).

PR #219, opened 2026-08-10
fix(ci): extract mergeable with jq tostring (#216)

### LINKED ISSUE #216: ci: consider explicit mergeable extraction (avoid @tsv null→empty ambiguity)
## Context

`check-pr-mergeable/check.sh` uses:

```bash
gh api ... --jq '[.mergeable, .mergeable_state] | @tsv'
```

JSON `null` becomes an empty TSV field. PR #213 / #214 fail closed on empty/non-true, which is correct, but the encoding remains ambiguous for operators reading raw fields.

## Proposed

- Prefer explicit stringification, e.g. `(.mergeable|tostring)` or dual-field JSON, so null is the literal `null` and mocks match production 1:1.
- Update mock harness and docs in the same change.

## Out of scope

- Broader Guard/Release workflow redesign
- Transient API retry policy (separate issue)

Related: #213, #214

MehdiZare:
## Summary

Closes #216 for the shared fail-closed mergeability action.

- Extract REST `.mergeable` with `jq tostring` so JSON `null` is the literal token `null` (not an empty `@tsv` field)
- Fail-closed contract unchanged: green only when `mergeable=true`
- Update mock matrix encoding locks (tostring null + bare-@tsv canary) and `action.yml` contract comments

## Why

Operators and mock fixtures reading raw TSV fields could not distinguish JSON null from an empty parse error. Explicit stringification keeps logs, `GITHUB_OUTPUT`, and mocks 1:1 with the REST field.

## Test plan

- [x] `bash .github/actions/check-pr-mergeable/test-check.sh` → 16 passed
- [ ] CI `actions-shell` job green
- [ ] Guard / Release still fail on dirty / false / null / unknown-after-retries; succeed only on `mergeable=true`

Closes #216


MehdiZare:
@coderabbitai review

MehdiZare:
## Review follow-up

Addressed stack review findings on this PR:

| Finding | Disposition |
|--------|-------------|
| `mergeable_state` JSON null could hit resolved `*)` and go green when `mergeable=true` | **Fixed** — map null state → `unknown` via `((.mergeable_state // "unknown") \| tostring)`; mock lock + `true-null-state-exhausts` |
| Bare `@tsv` canary long-term value | **Dropped** in favor of production extraction lock only |

Ready for merge once CI is green.

MehdiZare:
## Review / babysit update

- Rebased onto latest `dev` after **#220** merged (resolved `action.yml` / `CHANGELOG` combine with #218 docs).
- Added **defense-in-depth**: case arm treats literal `null` `mergeable_state` as unresolved (in addition to jq `// \"unknown\"`), with mock coverage feeding `true\\tnull`.
- Opened a follow-up issue for broader reliability/harness work (shared jq fixture, primary rate-limit retry, allowlisted success states) so this stack stays scoped.
- Stack order for auto-merge: **#219 → #221 → #222** (retarget base to `dev` after each lower PR merges).

Issue #252, opened 2026-08-13
Harden for production: MCP manifest execution, isolated publishing, and credential handling


MehdiZare:
## Summary

Independent security review of `9560a2e6` (current `HEAD` on `fix/ci-sync-auto-merge-and-signed-commit-friction`). I re-read the cited code, reproduced the high-risk behaviors locally, scanned the tree, and checked live GitHub security posture. **No code changes in this issue.** Use this as the tracker before any remediation PRs.

**Production verdict:** the core HTTPS client is in good shape for application use. The **MCP extra** and the **PyPI publish path** are not yet production-hardened. I would not call the package release pipeline ready against current GitHub/PyPI trusted-publisher guidance until the two High items are fixed.

| Severity | Count | Notes |
|---|---|---|
| Critical | 0 | No remotely exploitable unauthenticated RCE, no committed live credentials |
| High | 2 | MCP Python-manifest execution; publish job mixes OIDC/write with mutable inputs |
| Medium | 8 | 7 from the prior report (one partly overstated) + 1 new (Claude write-capable action) |
| Low | 3 | Path traversal, Makefile `.env` sourcing, missing `SECURITY.md` |

Prior report reviewed: `/tmp/fmp-data-security-best-practices-report.md`. This issue supersedes that report where the two disagree.

## What is already solid

- Default origin is HTTPS; TLS verify is left on.
- FMP HTTP error mapping redacts keys and suppresses `httpx.HTTPStatusError` chaining (`fmp_data/base.py` `_raise_fmp_http_error`).
- Cache keys drop `apikey`. File/Redis cache backends use JSON, not pickle.
- FAISS load is fail-closed (`allow_dangerous_deserialization=False`).
- No `shell=True` / `os.system` / `eval` / unsafe YAML in library source.
- File log handlers try mode `0600`; log dirs try `0700`.
- `main` ruleset: 2 approvals, Test-Matrix + CodeRabbit. `dev` ruleset: required checks, no force-push.
- Secret scanning + push protection enabled. Dependabot security updates enabled (weekly pip + Actions).
- Live APIs: **0** open Dependabot / code-scanning / secret-scanning alerts.
- `uv run pip-audit` on the current env: **no known vulns**. Bandit: **0** findings. Ruff: **pass**.
- `document.yml` already splits build vs deploy and uses `persist-credentials: false` — copy this pattern.
- TestPyPI PR + `dev` publish paths already use unique versions and `skip-existing: false` (the #204 class).

[...]

PR #255, opened 2026-08-13
fix(mcp): parse manifests as data instead of executing them

MehdiZare:
## Summary

Closes the High finding **FMP-SEC-001** from #252.

`fmp-mcp validate` and MCP server startup treated user-supplied `.py` manifests as importable modules (`spec.loader.exec_module`). A file that looked like a tool list could run arbitrary code in a process that already has `FMP_API_KEY` in the environment. The command name `validate` implied inspection.

Manifests are now **data**:

- Legacy `TOOLS = ["..."]` Python is parsed with a restricted AST (optional module docstring + that assignment only).
- Imports, calls, starred expressions, and any other statement are rejected. The file is never executed.
- JSON, YAML, and TOML are accepted as a top-level list or an object with a `tools` list.
- Existing generated and example `.py` manifests keep working unchanged.

A local probe of `open(...).write('RAN'); TOOLS = ['profile']` now raises instead of printing `Manifest is valid with 1 tool`.

## Test plan

- [x] New unit tests: executable Python is refused and leaves no side-effect file; imports/calls refused; JSON list + `{tools: []}` load; YAML loads; `validate` returns False without executing.
- [x] Existing example-manifest, generate/validate coherence, and CLI tests still pass (`220 passed` in the MCP unit files).
- [x] Pre-commit: ruff, bandit, mypy, detect-secrets.

## Compatibility

Not breaking for shipped manifests. Breaking only for anyone who used a `.py` manifest as a plugin (imports, function calls, `from ... import DEFAULT_TOOLS`). Those files were never a documented plugin API; they should become JSON/YAML or a plain `TOOLS = [...]` list.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -184,6 +184,15 @@ None. No FMP path we ship was newly retired by this changelog window.
   release**, set the PyPI and TestPyPI Trusted Publisher environment names
   to `pypi` and `testpypi` to match the workflows.
 
+- **MCP manifests are data, not executed code (#252).** `load_manifest_tools`
+  and `fmp-mcp validate` no longer `exec` user-supplied Python. Legacy
+  `TOOLS = ["..."]` files are parsed with a restricted AST (docstring + that
+  assignment only). Imports, calls, and any other statement are rejected.
+  JSON, YAML, and TOML manifests are also accepted (`["spec"]` or
+  `{"tools": ["spec"]}`). Existing generated and example `.py` manifests keep
+  working. A file that previously ran arbitrary code as "validation" now
+  fails closed.
+
 ## [2.6.0] - 2026-08-10
 
 Released from `dev`. A correctness-and-contracts release: the LangChain and MCP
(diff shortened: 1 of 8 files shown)


MehdiZare:
Reviewed without CodeRabbit (reviews are disabled on `dev`).

**Addressed in 0f1b613:**
- TOML is advertised on the 3.10 floor; it now uses `tomli` from the `mcp` extra (stdlib `tomllib` on 3.11+).
- CHANGELOG/docs no longer imply a top-level TOML array is valid.
- Tests lock a call and an f-string *inside* `TOOLS = [...]` (no side-effect file), plus `.toml`, `.yml`, malformed JSON `validate`, and a non-`TOOLS` assignment.

**Left as follow-ups (not this PR):**
- YAML/JSON size limits (local-file DoS, not RCE).
- `generate` still writes `.py`; opened a separate issue to emit JSON / honor the suffix.
- Remaining #252 items stay on that tracker.

Rebased onto `dev` after #254 so CHANGELOG keeps both security notes. Squash auto-merge is on.

Issue #320, opened 2026-08-15
P1: Setup wizard never offers example MCP profiles (wrong examples path)


MehdiZare:
## Why
`get_manifest_choices` in `fmp_data/mcp/utils.py` looks for manifests under:

```
examples/mcp_configurations/
```

The real directory is:

```
examples/mcp/configurations/
```

So the official `fmp-mcp setup` wizard only ever sees the default profile. Trading / minimal / research / crypto never appear.

The same stale path still appears in:

- `examples/mcp/claude_desktop/setup_claude_desktop.py`
- `examples/mcp/claude_desktop/README.md`
- `examples/mcp/claude_desktop/troubleshooting.md`

Docs (`docs/mcp/configurations.md`) were already corrected. There is even a test (`tests/unit/test_docs_endpoints_sync.py`) that caught the docs path but not the runtime helper.

## Proper fix (isolated PR)
1. Point `get_manifest_choices` at `examples/mcp/configurations/`.
2. Update the example Claude Desktop script and its README/troubleshooting copies.
3. Add a unit test that `get_manifest_choices()` returns the example profiles when those files exist.

Do not fold this into #315 (CodeQL logging). Isolated path fix.

Discovered while reviewing https://github.com/MehdiZare/fmp-data/pull/315

MehdiZare:
Fixed in #327. `get_manifest_choices` and the Claude Desktop examples (including the advanced JSON) use `examples/mcp/configurations/`. Merged to `dev` as 0e50b41.

Issue #319, opened 2026-08-15
P1: Sanitize MCP helper stderr at the source so CLI cannot print keys


MehdiZare:
## Why
#315 stopped the setup wizard from echoing `validate_api_key` / `test_mcp_server` messages because those strings can include subprocess stderr that quotes the API key. CodeQL treats `print` / stdout as a logging sink.

The same helpers are still printed verbatim by the CLI:

- `fmp_data/mcp/cli.py` `status_command` (`print(f"❌ {message}")`)
- `fmp_data/mcp/cli.py` `test_command` (same)

`test_mcp_server` returns `f"MCP server test failed: {error_msg}"` with raw stderr. That is the same class of leak #315 closed in the wizard.

## Proper fix (isolated PR)
Return classified, key-free reasons from `fmp_data/mcp/utils.py` (`validate_api_key`, `test_mcp_server`) — e.g. an enum / literal (`invalid`, `timeout`, `unavailable`) plus a static display string.

Stop embedding `{error_msg}` / `{e}` in the returned message. Then both the wizard and the CLI can print the classified string safely.

Do not paper this over with more `# codeql[…]` suppressions.

## Out of scope
Do not fold this into #315. The wizard already uses static / allowlisted copy.

Discovered while reviewing https://github.com/MehdiZare/fmp-data/pull/315

MehdiZare:
Review follow-up from #315: allowlisting the helper's return string is **not** enough. CodeQL still treats `status if status in SAFE else "…"` as dataflow from getpass and flags `stdout.write`. The classified string has to be a **literal constructed at the sink** (or a typed reason enum mapped to literals), not the helper's own return value. #315 now prints only static copy; this issue is the proper fix.

MehdiZare:
The #315 squash shipped the allowlist echo. Isolated revert-to-literals is #321. This issue remains the proper classified-reason design.

MehdiZare:
Fixed in #327. Helpers return classified reasons; `fmp-mcp status`/`test` print sink-local literals. Merged to `dev` as 0e50b41.

Issue #318, opened 2026-08-15
P3: CodeRabbit auto-review is skipped for PRs targeting dev


MehdiZare:
## Why
PRs that target \`dev\` (including `#315`) get:

> Review skipped. Auto reviews are disabled on base/target branches other than the default branch.

Default branch is \`main\`. Most work lands on \`dev\` first. That leaves CodeRabbit as a no-op on the branch where we actually review.

## Proper fix (isolated PR / config change)
Update \`.coderabbit.yaml\` so auto-review also runs for \`dev\` (CodeRabbit \`reviews.auto_review.base_branches\`, or the current equivalent). Keep draft skipping.

Do not block `#315` on this.

Discovered while reviewing https://github.com/MehdiZare/fmp-data/pull/315

MehdiZare:
Fixed in #326 (`reviews.auto_review.base_branches: ["^dev$"]`). Merged to `dev` as 0b7f665.

Issue #317, opened 2026-08-15
P2: validate_api_key only constructs a client — it never calls FMP


MehdiZare:
## Why
`fmp_data/mcp/utils.py` `validate_api_key` runs:

```
FMPDataClient.from_env(); client.close()
```

`from_env` / `ClientConfig.validate_api_key` reject empty or all-\`*\` keys, not a 401. A typed junk key still produces \"API key validated successfully.\" in the setup wizard.

The helper is also used as the wizard's validation step, so users can persist a bad key into Claude Desktop config.

## Proper fix (isolated PR)
Make a cheap authenticated probe (e.g. a stable, low-cost FMP endpoint) and map 401/403 to the existing classified string \`API key is invalid or expired\`. Keep stderr off the returned message (see the P1 helper-sanitization issue).

Do not fold this into `#315`.

Discovered while reviewing https://github.com/MehdiZare/fmp-data/pull/315

MehdiZare:
Fixed in #327. `validate_api_key` now probes FMP via `company.get_quote("AAPL")`. 401/403 → invalid. Merged to `dev` as 0e50b41.

PR #328, opened 2026-08-15
fix(security): unify credential display redaction behind two APIs

### LINKED ISSUE #316: P2: Unify credential display redaction (setup / base / _redaction)
## Why
Three independent redaction implementations now exist:

- `fmp_data/mcp/setup.py` `_redact_key_patterns` (console, pattern-only — must not read the getpass-held key; CodeQL logging sink)
- `fmp_data/base.py` `_redact_api_keys` (query-string / assignment / percent-encoded)
- `fmp_data/_redaction.py` (key-name walk for config `__str__` / embeddings)

They will drift. `#315` had to duplicate URL + key-shaped patterns in the wizard because routing console output through a helper that reads the held secret re-triggers `py/clear-text-logging-sensitive-data`.

## Proper fix (isolated PR)
One module with two explicit APIs:

1. **Pattern-only** (CodeQL-safe for logging sinks / stdout): never takes the live secret as an argument.
2. **Exact-replace** (prompts / `input` / `getpass` only): may take the held secret.

Call sites choose by sink type. Add a CodeQL regression test or a comment pointing at the query so we do not re-taint stdout.

Out of scope for `#315`.

Discovered while reviewing https://github.com/MehdiZare/fmp-data/pull/315

MehdiZare:
## Why

Three independent redaction implementations existed after #315:

- `fmp_data/mcp/setup.py` `_redact_key_patterns` (console, pattern-only)
- `fmp_data/base.py` `_redact_api_keys` (query-string / assignment / percent-encoded)
- `fmp_data/_redaction.py` (key-name walk for config `__str__` / embeddings)

They will drift. Routing console output through a helper that reads the held secret re-triggers CodeQL `py/clear-text-logging-sensitive-data`.

Closes #316

## What changed

One module, two sink-appropriate APIs, plus a wizard-only extra:

1. **`redact_credential_patterns(text)`** — query / assignment / encoded. Never takes the live secret. Used by HTTP error bodies and the wizard.
2. **`redact_held_secret(text, secret)`** — exact replace. Prompts / `getpass` only.
3. **`redact_key_shaped_tokens(text)`** — `sk-` / 32-char heuristics. Wizard console only. Not composed into `base._redact_api_keys` so request ids are not blanked.

`base._redact_api_keys` and the setup wizard both call through the shared assignment/URL rules so those cannot drift.

## Out of scope

#317 / #319 / #320 stay on #327 (classified helper reasons, live FMP probe, example-profile path). This PR previously bundled a weaker copy of that work; it was reverted so the redaction unification stays isolated.

#318 (CodeRabbit auto-review on `dev`) stays on #326.

`closes #N` is inert on `dev`-base PRs; close the issue by hand after merge if the link does not fire.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,12 +10,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - **One module now owns credential display redaction (#316).**
-  `fmp_data._redaction` exposes two explicit APIs: pattern-only
-  `redact_credential_patterns` (CodeQL-safe for stdout / logging sinks;
-  never takes the live secret) and exact-replace `redact_held_secret`
-  (prompts / `getpass` only). `base._redact_api_keys` and the MCP setup
-  wizard both call through those, so the three former copies cannot
-  drift.
+  `redact_credential_patterns` is the shared query/assignment/encoded
+  path (CodeQL-safe for stdout / logging sinks; never takes the live
+  secret). `redact_held_secret` is exact-replace for prompts /
+  `getpass` only. The setup wizard also runs `redact_key_shaped_tokens`
+  (sk-/32-char heuristics); those stay off HTTP error bodies so request
+  ids are not blanked.
 
 ## [2.7.0] - 2026-08-14
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   wizard both call through those, so the three former copies cannot
   drift.
 
+- **MCP helper messages are classified and key-free (#319).**
+  `validate_api_key` and `test_mcp_server` return only static display
+  strings (`API_KEY_*`, `MCP_SERVER_*`). They no longer embed stderr or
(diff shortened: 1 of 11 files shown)


MehdiZare:
## Review and scope isolation

This PR had two commits. The second (`eeacc5e`) re-implemented #317 / #319 / #320, which already live on #327 with the stronger CLI sink-local design (`_print_server_check`). Allowlisted helper display strings are still CodeQL-tainted (#321). That work does not belong here.

**Done:**
- Reverted the overlapping MCP helper commit. #317 / #319 / #320 stay on #327.
- Split wizard key-shaped heuristics (`sk-` / 32-char / hex-40) out of `redact_credential_patterns`. `base._redact_api_keys` now only runs query/assignment/encoded rules. A 32-char request id in an HTTP error body is no longer blanked.
- Tests pin both contracts.

CI is re-running on `1814166`.

MehdiZare:
Review wrap-up: over-redaction split is on `1814166`, overlapping MCP work stays on #327.

Leftover: the assignment lookbehind that preserves `export FMP_API_KEY="[YOUR_API_KEY]"` also skips `FMP_API_KEY=` in HTTP error bodies. Isolated follow-up: #330 (P3).

CI is green. Squash-merging.

PR #331, opened 2026-08-15
fix(mcp): treat status-less Invalid API KEY as invalid

### LINKED ISSUE #329: P3: validate_api_key treats 200-body "Invalid API KEY" as valid
## Why

`#317` / `#327` classify a live probe as `invalid` on `AuthenticationError` and on `FMPError.status_code in {401, 403}`.

`base._check_error_response` still raises a bare `FMPError(message)` with **no** `status_code` when a **200** body looks like:

```json
{"Error Message": "Invalid API KEY"}
```

That path is treated as `valid` (key accepted) because `status_code is None`. If stable `/quote` still returns this shape for a junk key, the wizard will persist it — the same hole `#317` closed for HTTP 401/403.

## Isolated fix

In `validate_api_key` (or in `_check_error_response` if we want it everywhere), treat a status-less `FMPError` whose message matches the known invalid-key copy as `invalid`. Do not over-match unrelated 200 error bodies.

Add a unit test that `FMPError("Invalid API KEY")` (no status) → `(False, "invalid")`.

Discovered while reviewing #327.

MehdiZare:
Closes #329.

## Probe (2026-08-15)

The issue described a **200** body `{"Error Message": "Invalid API KEY"}` being treated as a valid key because `status_code is None`.

Live FMP no longer does that on the probe path `validate_api_key` uses:

| Request | Status | Body |
|---|---|---|
| `GET /stable/quote?symbol=AAPL&apikey=junk` | **401** | `{"Error Message": "Invalid API KEY. Feel free to create a Free API Key..."}` |
| `GET /api/v3/quote/AAPL?apikey=junk` | **401** | same |

HTTP 401 already becomes `AuthenticationError` in `handle_response` → `invalid`. The 200-body hole is not live on `/quote` today.

## What this PR still changes

`base._check_error_response` still raises a bare `FMPError(message)` with **no** `status_code` if a 2xx JSON body is that copy (other endpoints, or a FMP regression). `validate_api_key` now treats that known copy as `invalid`.

Unrelated status-less `FMPError`s (e.g. `"Legacy endpoint retired"`) still count as `valid`. 429 / 5xx still count as `valid`.

## Tests

- `FMPError("Invalid API KEY")` (no status) → `(False, "invalid")`
- status-less unrelated `FMPError` stays `(True, "valid")`

## Out of scope

Did not change `_check_error_response` globally (user picked isolated, wizard-local fix after the probe).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved API-key validation for responses without a status code.
  * Errors containing “Invalid API KEY” are now correctly classified as invalid.
  * Other status-less errors continue to be treated as valid.

* **Documentation**
  * Added a changelog entry describing the updated API-key validation behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,6 +55,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   client and calls `company.get_quote("AAPL")`; 401 and 403 map to
   `invalid`. Rate-limit / other HTTP errors still count as `valid`.
 
+- **`validate_api_key` treats a status-less `Invalid API KEY` body as
+  invalid (#329).** Live `/stable/quote` (and `/api/v3/quote`) now
+  return HTTP 401 for a junk key (probed 2026-08-15), which already
+  maps to `AuthenticationError`. `_check_error_response` still raises
+  a bare `FMPError` with no `status_code` if a 2xx JSON body is
+  `{"Error Message": "Invalid API KEY..."}`; that copy now classifies
+  as `invalid`. Unrelated status-less `FMPError`s still count as
+  `valid`.
+
 - **Setup wizard offers the example MCP profiles again (#320).**
   `get_manifest_choices` looks in `examples/mcp/configurations/` (the
   real directory). The example Claude Desktop script, JSON profiles,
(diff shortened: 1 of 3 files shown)


MehdiZare:
## Review follow-up

Comprehensive review on this branch (CodeRabbit had no actionable inline comments). No unresolved review threads.

**Addressed in this push**
- Parametrized the unit test so the live copy (`Invalid API KEY. Feel free to create a Free API Key...`) stays `invalid`, not only the short issue example.
- Trimmed the design-history comment above the matcher.
- Rebased onto current `dev` (strict status checks require an up-to-date branch).

**Follow-up (out of scope here)**
- #340 (P2): raise `AuthenticationError` from `_check_error_response` for that copy so every caller gets a typed auth failure, not only `validate_api_key`. Isolated on purpose after the #329 probe.

Issue #352, opened 2026-08-17
[P3] Promote Structure to the public fmp_data export surface


MehdiZare:
Follow-up from #351 / #349. Not a merge blocker.

`Structure = Literal["flat", "nested"]` is now a closed request vocabulary on public methods (`get_product_revenue_segmentation` / `get_geographic_revenue_segmentation`, sync and async). The other closed vocabs (`Period`, `PeriodAnnualQuarter`, `PeriodFiscal`, `Interval`, `Timeframe`, `TechnicalInterval`) are re-exported from `fmp_data` and pinned by `test_closed_vocabularies_are_public_exports`.

#351 deliberately left `Structure` off `__all__` — same deferred decision as #308 / #311 before Period / TechnicalInterval landed. The negative pin (`"Structure" not in fmp_data.__all__`) is there so this cannot drift silently.

## Isolated next steps

1. Import `Structure` in `fmp_data/__init__.py` and add it to `__all__`.
2. Flip the negative pin in `test_closed_vocabularies_are_public_exports` to `assert fmp_data.Structure is Structure` / `"Structure" in fmp_data.__all__`.
3. Changelog Unreleased / Added, same voice as #311.

Do not fold this into a catalogue sweep. `StructureTypeEnum` stays leftover / deprecated.

Optional extra (only if it stays small): `STRUCTURE_HINT` / the endpoint `description` still say "Data structure format" and do not mention that live `/stable` currently returns the same list-of-objects for `flat` and `nested` (probed 2026-08-17). Method docstrings already say this; LangChain tools read the endpoint description.

Related: #351, #349, #308, #311.

MehdiZare:
Implementation in #356 (isolated from #354).

`Structure` is now re-exported from `fmp_data`. The #349 negative pin is flipped. Endpoint descriptions and `STRUCTURE_HINT` mention that live `/stable` returns the same list-of-objects for `flat` and `nested`. `StructureTypeEnum` stays leftover.

`Closes #352` is inert on a `dev`-base PR — close this by hand after #356 merges.

MehdiZare:
Closed by #356 (`feat(schema): re-export Structure from the public fmp_data surface`). `Structure` is now on `fmp_data.__all__` next to `Period` / `TechnicalInterval`. `StructureTypeEnum` stays leftover / deprecated. `Closes #352` is inert on a `dev`-base PR, so this is the hand close.

Issue #94, opened 2026-04-13
[security] Redact API key from httpx error tracebacks


MehdiZare:
## Summary

`fmp-data` can leak the FMP API key in Python tracebacks when an HTTP error response is raised by `httpx`.

The top-level `fmp_data` exception message is mostly clean, but the original `httpx.HTTPStatusError` remains in the exception chain via `raise ... from error`. `httpx` includes the full request URL in that exception message, and `fmp-data` currently sends the key as an `apikey` query parameter.

This was found while investigating `bina-capital` issue MehdiZare/bina-capital#376. That issue contains a real leaked key, so do not copy any traceback content from there into this repository.

## Repro

Use a fake key only:

```python
import traceback

import httpx
from fmp_data import ClientConfig
from fmp_data.base import BaseClient
from fmp_data.exceptions import RateLimitError

secret = "SECRET_FMP_KEY_123"
client = BaseClient(ClientConfig(api_key=secret, max_retries=1, max_rate_limit_retries=0))
request = httpx.Request(
    "GET",
    f"https://financialmodelingprep.com/stable/quote?symbol=AAPL&apikey={secret}",
)
response = httpx.Response(429, request=request, json={"error": "limit"})

try:
    client.handle_response(response)
except RateLimitError:
    tb = traceback.format_exc()
    assert "apikey=" in tb
    assert secret in tb
finally:
    client.close()
```

Observed traceback includes a line shaped like:

```text
httpx.HTTPStatusError: Client error '429 Too Many Requests' for url 'https://financialmodelingprep.com/stable/quote?symbol=AAPL&apikey=SECRET_FMP_KEY_123'
```

## Suspected Root Cause

In `fmp_data/base.py`:

- `_execute_request()` and async request execution add `query_params["apikey"] = self.config.api_key` before calling `httpx`.
- `handle_response()` calls `response.raise_for_status()`.
- `_handle_http_status_error()` raises typed `fmp_data` exceptions using `raise ... from error`, preserving the original `httpx.HTTPStatusError` as the traceback cause.

## Acceptance Criteria

- Tracebacks for 4xx/5xx errors must not contain `apikey=` or the configured key value.
- Add unit coverage for at least a 429 response that asserts both the traceback and `str(exc)` are redacted.
- Cover both sync and async paths, or move shared sanitization into `handle_response()` / `_handle_http_status_error()` so both paths use the same behavior.
[...]

PR #353, opened 2026-08-18
fix(sec): map timeout and network errors without leaking apikey (#350)

### LINKED ISSUE #350: [security] Timeout / network httpx errors still leak apikey via URL and __cause__
## Why

`#94` / `#97` closed this for **HTTP status** errors: typed `FMPError`s no longer chain the raw `httpx.HTTPStatusError`, so a 429/401 traceback no longer carries `apikey=`.

The same class is still open on two other paths. We just reproduced it downstream in [bina-capital#2610](https://github.com/MehdiZare/bina-capital/issues/2610).

FMP authenticates with `apikey=` as a **query parameter**. httpx puts the full request URL in `str(exc)` and on `.request.url`. A clean typed message is not enough — Sentry serialises `__cause__` / `__context__` and `exc_info=True` logs.

## Status

Fixed in PR #353 (merged to `dev`). Leftover `httpx.RequestError` subclasses are **#354** (P1).

## Path 1 — still in this repo

`_execute_request` (sync and the async twin) still does, on `dev`:

```python
except Exception as e:
    self.logger.error(
        f"Request failed: {e!s}",
        extra={"endpoint": endpoint.name, "error": str(e)},
        exc_info=True,
    )
    raise
```

A timeout or connection error never goes through `handle_response` / `_handle_http_status_error`. The raw `httpx.TimeoutException` / `httpx.NetworkError` is logged (`str(e)` **and** `exc_info=True`) and re-raised. Those strings *are* the request URL.

`query_params["apikey"] = self.config.api_key` is still how the key travels (header was correctly removed on `dev` for `#252`). So this path still plants a live credential in logs and in any consumer that wraps the exception.

`#94`'s acceptance criteria said 4xx/5xx. Timeouts were not in that pin.

### Repro (fake key only)

```python
import httpx
from fmp_data import ClientConfig
from fmp_data.base import BaseClient

secret = "SECRET_FMP_KEY_123"
# Point the client at a black-hole so the request times out.
client = BaseClient(
    ClientConfig(api_key=secret, timeout=0.01, max_retries=1, max_rate_limit_retries=0)
)
try:
    # any endpoint; the transport never answers
    client.client.get(
        "https://10.255.255.1/stable/quote",
        params={"symbol": "AAPL", "apikey": secret},
    )
except httpx.HTTPError as exc:
    assert secret in str(exc) or (exc.request is not None and secret in str(exc.request.url))
finally:
    client.close()
```

Same shape we measured on the consumer side:

```
message      : <path> HTTP …          # often clean
__cause__    : HTTPStatusError / TimeoutException
cause.request.url contains key: True
```

## Path 2 — consumers reintroduce `#94` the moment they wrap httpx

`bina-fmp` (the Bina adapter) has first-party httpx clients for surfaces this package does not own yet (`/stable/historical-price-eod/dividend-adjusted`, PIT statements, a few `/stable/*` AI-map calls). Those do:

```python
raise FMPStablePriceError(f"{path} HTTP {exc.response.status_code}") from exc
```

`str(exc)` is clean. `__cause__.request.url` is not. That is exactly `#94`, one layer up.

We will fix the adapter. Filing here because:

1. The timeout/network path above is the same defect still in-tree.
2. `_redact_api_keys` / `redact_credential_patterns` is private. Consumers cannot reuse it, so every first-party wrapper re-implements a weaker copy (message-only, leave `__cause__`).
3. A test that only asserts `secret not in str(exc)` goes green on the leaking state. `#97` already knew this; the consumer tests did not.

## Scope

- Map `httpx.TimeoutException` / `httpx.NetworkError` to a typed `FMPError` / `FMPError` subclass with `raise … from None`. Do not log `str(httpx_exc)` or `exc_info=True` on that path.
- Pin a timeout (or MockTransport that raises `httpx.ConnectError` whose request URL contains `apikey=`) and assert the secret is in **neither** `str(exc)` **nor** `repr(exc.__cause__)` **nor** `traceback.format_exc()`.
- Same pin on the async client.
- Consider exporting the redaction helper (or a `raise_secret_safe(...)`) so adapters that must speak httpx themselves cannot quietly re-open `#94`.

## Out of scope

- `SecretStr` on `ClientConfig.api_key` (#252 / 2.7.0) — necessary, not sufficient. A `SecretStr` still becomes a query string; httpx will still put it in the URL.
- Adding every `/stable` surface bina-fmp currently wraps. Nice if you cover dividend-adjusted EOD; not required to close this.

## Related

- This repo: `#94` (closed), `#97` (status-error `from None`), `#252` (SecretStr + no header key), `#353` (this fix), `#354` (remaining RequestError leak)
- Downstream: [MehdiZare/bina-capital#2610](https://github.com/MehdiZare/bina-capital/issues/2610)

MehdiZare:
## Summary

`#94` / `#97` closed this for HTTP **status** errors. Timeouts and connect failures never reached `handle_response`: sync and async `_execute_request` logged `str(httpx_exc)` with `exc_info=True` and re-raised raw `httpx.TimeoutException` / `httpx.NetworkError`. httpx puts the request URL on those objects; the key is a query param. Sentry serialises `__cause__` / `.request.url`.

This PR:

- Maps `httpx.TimeoutException` → `FMPTimeoutError` and `httpx.NetworkError` → `FMPNetworkError` with `raise … from None`.
- Logs a fixed line (`Request timed out` / `Network error`) with `endpoint` extra only. No `str(httpx_exc)`, no `exc_info=True` on that path.
- Keeps those failures retryable (`_is_retryable_error`).
- Re-exports both types from `fmp_data` next to `RateLimitError`.
- Pins the secret is absent from `str(exc)`, `repr(exc)`, `repr(exc.__cause__)`, `traceback.format_exc()`, and the error-log call. Sync and async.
- Classifies those mapped types in `validate_api_key`: `FMPTimeoutError` → `timeout`, `FMPNetworkError` → `unavailable`. Without that, the generic `except FMPError` branch (`status_code is None`) would report a valid key.

Callers that caught raw httpx timeouts should catch the new types (they remain `FMPError` subclasses).

Not in this PR:

- Public `raise_secret_safe` / exporting `_redaction` (bina-capital#2610 / PR #2690 handles the adapter).
- `#352` Structure export.
- Remaining `httpx.RequestError` subclasses (`ProtocolError`, `ProxyError`, …) — follow-up **#354** (P1).

`Closes #350` is inert on a `dev`-base PR — close #350 by hand after merge.

## Test Plan

- [x] `uv run pytest` focused suite (`test_base`, `test_exceptions`, `test_mcp_utils`, leak pins, `test_retry_on_timeout`)
- [x] `ruff` + `mypy` on touched modules; pre-commit (ruff, bandit, detect-secrets, mypy) passed on commit
- [ ] Protect Dev / Test-Matrix on this PR
- [ ] Close #350 by hand after merge


MehdiZare:
## Review (CodeRabbit rate-limited)

Did not wait on CodeRabbit. Review of `800655d` vs `origin/dev`:

### Addressed in this PR
`validate_api_key` still caught only `httpx.TimeoutException`. After the client maps transport failures to `FMPTimeoutError` / `FMPNetworkError` (`FMPError`, `status_code is None`), the generic `except FMPError` branch treated a timeout or connect failure as a **valid key**. That is a first-party caller of this mapping, so it stayed in-scope as an isolated fix (`800655d`). Pins:

- `FMPTimeoutError` → `timeout`
- `FMPNetworkError` → `unavailable`
- existing fake `httpx.TimeoutException` path left in place

### Follow-up, not expanded here
Remaining `httpx.RequestError` subclasses (`ProtocolError`, `ProxyError`, `TooManyRedirects`, …) still hit the generic logger (`str(exc)` + `exc_info=True`) and re-raise raw httpx. Same leak class, broader type/retry design. Tracked as **#354** (P1).

### Otherwise
Mapping + `from None`, retryability, public re-exports, and the #97-shaped secret pins (sync + async) look correct. No unresolved review threads.

Auto-merge (squash) will be enabled once Protect Dev checks are green.

PR #373, opened 2026-08-19
Sync: main → dev (history reachability)

MehdiZare:
## Automatic Sync

`main` was not an ancestor of `dev`. This PR restores reachability so
the next `dev → main` release PR stays MERGEABLE and gets full CI.

### Why this is not a content sync
After a squash-merge release, trees are usually identical but histories
diverge. Comparing content (`git diff`) misses that; comparing
reachability (`git merge-base --is-ancestor`) catches it at the right
moment — right after the release — instead of as 20+ conflicts on the
next release PR. See #202.

### Merge policy
- Auto-merge is enabled with a **merge commit** when required checks
  (Test-Matrix) are green. Do **not** squash — that re-creates the
  ancestry break this PR fixes.
- Content is usually empty for a pure history merge; review only if
  the diff is non-empty (unexpected).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,7 +36,7 @@ jobs:
           python-version: ${{ matrix.python }}
 
       - name: ⚡  Install uv (with cache)
-        uses: astral-sh/setup-uv@ae62891fec2bb8e7d6c99fc78c9fec3a63790f8d  # v10.0.0
+        uses: astral-sh/setup-uv@20cfd1bf945f4377ade1205e4dbc17946fc9a30d  # v10.0.1
         with:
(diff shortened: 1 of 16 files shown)


MehdiZare:
**Checked.** History-only `-s ours` merge (`d007a22`, parents `37573a6` + `6f66718`). Trees of `dev`/`main`/`sync` were identical (`cc85e10`); file diff empty. Auto-merge used a **merge commit** (not squash).

Merged as `55b80fa`. `main` is an ancestor of `dev` again. Test-Matrix was green. No review threads.

Do **not** squash-merge the new standing release PR #374 until there is a real content delta — that would re-break ancestry for an empty tree.

Issue #370, opened 2026-08-19
[P3] Rewrite published GitHub Release 2.7.0 notes to match in-repo CHANGELOG


MehdiZare:
Raised while reviewing #369 (the #360 leftover).

#369 rewrote the in-repo `## [2.7.0]` section to one Added / Changed / Fixed / Security with `### FMP API surface` first. The published GitHub Release for `v2.7.0` was left as the generated stub on purpose — editing a shipped release is a separate product decision, not a changelog-file fix.

**Wanted:** either
1. Replace the GitHub Release body with the in-repo 2.7.0 section (minus Unreleased), or
2. Document that GitHub Release notes are generated from the squash commit list and are not kept in sync with `CHANGELOG.md`.

Out of scope for #369. Not a functional defect.

Related: #360, #369, #304.

MehdiZare:
Opened from the #369 review. In-repo 2.7.0 rewrite is in #369 (`0d65f6d` + the original leftover commit). This issue is only the published GitHub Release body.

MehdiZare:
Two parts:

1. Live `v2.7.0` GitHub Release body rewritten from the in-repo `## [2.7.0]` section: https://github.com/MehdiZare/fmp-data/releases/tag/v2.7.0
2. Generator change so the next cut does not fall back to the squash-commit stub: https://github.com/MehdiZare/fmp-data/pull/379 (`scripts/github_release_notes.py`, `release.yml` fail-closed if the CHANGELOG section is missing).

MehdiZare:
Landed in #379 (`2da191d` on `dev`).

The ticket was stale: the published [v2.7.0](https://github.com/MehdiZare/fmp-data/releases/tag/v2.7.0) body already matched `CHANGELOG.md`. Future releases extract the matching `## [X.Y.Z]` section via `scripts/github_release_notes.py` (fail-closed if missing/empty) instead of the squash commit list. Older GitHub Releases (v2.6.0 and earlier) were left as squash stubs on purpose.

GitHub did not auto-close this because the default branch is `main`.

PR #382, opened 2026-08-19
fix(release): generate CHANGELOG notes before tagging (#370)

MehdiZare:
## Summary

Isolated leftover of #370 / review of [PR #380](https://github.com/MehdiZare/fmp-data/pull/380).

`release.yml` extracted CHANGELOG notes *after* pushing the annotated tag. A missing `## [X.Y.Z]` (the next labeled cut, with notes still under Unreleased) would then fail closed with a remote tag and no GitHub Release / PyPI publish. Re-runs hit “tag already exists.”

- Move generate-notes to immediately after version calculation (Python is already set up).
- Pin argv (`--changelog`, `--version`, `--tag`, `--out`), env (`VERSION` / `VERSION_NUMBER`), empty-file guard, and step order.
- CLI output must equal `render(extract(...))`.
- `docs/contributing/releasing.md` no longer describes git-log autogen.

Closes nothing on its own; keeps #370’s fail-closed policy from leaving orphan tags.

## Test plan

- [x] `uv run pytest tests/unit/test_release_tag_push.py tests/unit/test_github_release_notes.py tests/unit/test_changelog_headings.py tests/unit/test_workflow_run_scripts.py` (29 passed)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -134,6 +134,22 @@ jobs:
             echo "VERSION_NUMBER=$NEW_VERSION"
           } >> "$GITHUB_OUTPUT"
 
+      - name: 📋 Generate release notes from CHANGELOG
+        env:
+          RELEASE_TAG: ${{ steps.version.outputs.VERSION }}
+          RELEASE_VERSION: ${{ steps.version.outputs.VERSION_NUMBER }}
+        run: |
+          set -euo pipefail
+          # Body is the matching ## [X.Y.Z] section, not the squash commit
+          # list. Missing / empty section fails the job *before* the tag is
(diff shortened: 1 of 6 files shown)


MehdiZare:
## Review

Isolated leftover of #370 from the #380 review: generate CHANGELOG notes before creating/pushing the tag so a missing `## [X.Y.Z]` cannot leave an orphan remote tag.

Argv/env/step-order pins and `render(extract(...))` equality are in the tests. Docs no longer describe git-log autogen.

Folding Unreleased into the next version heading is [#383](https://github.com/MehdiZare/fmp-data/issues/383) (P2) — do that before any `release:*` label.

